PR #1, opened 2015-10-20
Avoid using _unused

Lacaranian:
Use the result information to catch bad commands before running them


Lacaranian:
(this is probably more a matter of style so far, since readers tend to be pretty safe. Let me know what you think!)

mgattozzi:
I prefer this usage of error handling and it's more verbose error wise so I merged the commit. Thanks!

PR #2, opened 2015-12-15
Sync ion up with shell from redox repository

skylerberg:
Please review this change before merging. There are several things to note:

* This pull request deletes the original code in the repository.
* This pull request merges the shell from the redox repo preserving as much history as possible.
* The code does not compile because of `#[no_mangle]`.
* For a time there will be two copies of the shell.

Things that still need to be done after this is merged:
* This repo needs to be included in the main repo instead of being duplicated.
* This needs to compile by not using `#[no_mangle]`.
* To make this work in the main repo, a shim with `#[no_mangle]` needs to be included.

Ion should be added as a submodule in the `crates` directory in redox, in my opinion, but I don't know if this will automatically be linked so that we can add the shim application into `filesystem/apps/shell`.


jackpot51:
Ok, here goes

PR #6, opened 2015-12-17
Switch to maps for variables and commands

skylerberg:
Environment variables and commands are now stored in maps. This simplifies the code and makes it more efficient. In the future we should switch to HashMaps, but BTreeMaps will do just fine for now.


mgattozzi:
Thanks for the change! I'll merge it

PR #8, opened 2015-12-17
Handle multiple commands per line; move toward a parser

skylerberg:
Details in commit messages.

### DIFF
--- a/src/expansion.rs
+++ b/src/expansion.rs
@@ -1,0 +1,25 @@
+use std::collections::BTreeMap;
+use super::tokenizer::Token;
+
+pub fn expand_tokens(tokens: &mut Vec<Token>, variables: &BTreeMap<String, String>) -> Vec<Token> {
+    let mut expanded_tokens: Vec<Token> = vec![];
+    for token in tokens.drain(..) {
+        expanded_tokens.push(match token {
+            Token::Word(word) => {
+                if word.starts_with('$') {
+                    let key = word[1..word.len()].to_string();
+                    if let Some(value) = variables.get(&key) {
+                        Token::Word(value.clone())
+                    }
+                    else {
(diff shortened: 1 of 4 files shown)


jackpot51:
Make sure you have lots of tests! Shellshock was added to bash at around this stage

Ticki:
We actually need better testing a lot of places.

skylerberg:
I have added quite a few tests so far. I am trying to break the large untested main file into modular parts that are each tested.

jackpot51:
The tests here are very good so far, thanks

src/expansion.rs
```diff
@@ -0,0 +1,25 @@
+use std::collections::BTreeMap;
+use super::tokenizer::Token;
+
+pub fn expand_tokens(tokens: &mut Vec<Token>, variables: &BTreeMap<String, String>) -> Vec<Token> {
```
hauleth: You should use `&mut [T]` instead of `&mut Vec<T>`.

hauleth:
Why use custom parser instead of existing ones like `nom`?

skylerberg:
I hadn't heard of `nom`. It looks neat! I will take a closer look and hopefully we can use it.

Ticki:
Isnt nom overkill for this task?

hauleth:
We are creating language here. Why writing your own buggy code when someone can do this for you?

Ticki:
Yeah, you're right. Nom is fine for this task.

hauleth:
There are also other solutions like [Lalrpop](https://github.com/nikomatsakis/lalrpop), but I find Nom simplest and easiest.

Ticki:
@hauleth Lalrpop is LR(1), but nom is not. I doubt Lalrpop will be good for this task (if we want a non-trivial language).

PR #18, opened 2016-01-11
Add formatting check to .travis.yml

polymetric1:


.travis.yml
```diff
@@ -2,6 +2,15 @@ language: rust
 rust:
   - nightly
 sudo: false
+before_script:
```
skylerberg: You need to create the formatted folder first. `mkdir formatted`.

skylerberg:
See inline comment.

In general this is not a big issue, but if you want to make your pull requests prettier you can do an interactive rebase and squash commits that do not add to the history. For example " attempt to fix cp command again" could be squashed.

polymetric1:
Yeah, I will definitely do that. I keep recommitting to test the travis thing, but I realize i shouldn't have to do that. I'm still trying to figure out a way to get the `cp` command to work

skylerberg:
the problem is that `formatted` does not exist.

You need to do
```
mkdir formatted
cp -r * formatted/

polymetric1:
I know, but that still causes an error and halts the build.

polymetric1:
I'll still try it, though, I just haven't had a chance to

polymetric1:
I'm going to close this PR and create a new one when I have this figured out.

PR #23, opened 2016-01-16
Move builtins to their own module

skylerberg:
This change should make it possible to test builtins easier. Furthermore builtin functions now have function signatures that reflect how they actually operate.

### DIFF
--- a/src/builtin.rs
+++ b/src/builtin.rs
@@ -11,7 +11,7 @@ use super::to_num::ToNum;
 use super::set_var;
 use super::input_editor::readln;
 
-pub fn cat(args: &Vec<String>) {
+pub fn cat(args: &[String]) {
     let path = args.get(1).map_or(String::new(), |arg| arg.clone());
 
     match File::open(&path) {
@@ -26,7 +26,7 @@ pub fn cat(args: &Vec<String>) {
     }
 }
 
-pub fn cd(args: &Vec<String>) {
+pub fn cd(args: &[String]) {
     match args.get(1) {
         Some(path) => {
             if let Err(err) = env::set_current_dir(&path) {
@@ -37,7 +37,7 @@ pub fn cd(args: &Vec<String>) {
     }
 }
 
-pub fn echo(args: &Vec<String>) {
+pub fn echo(args: &[String]) {
     let echo = args.iter()
                    .skip(1)
                    .fold(String::new(), |string, arg| string + " " + arg);
@@ -57,7 +57,7 @@ pub fn free() {
     }
 }
 
-pub fn ls(args: &Vec<String>) {
+pub fn ls(args: &[String]) {
     let path = args.get(1).map_or(".".to_string(), |arg| arg.clone());
 
     let mut entries = Vec::new();
@@ -99,7 +99,7 @@ pub fn ls(args: &Vec<String>) {
     }
 }
 
-pub fn mkdir(args: &Vec<String>) {
+pub fn mkdir(args: &[String]) {
     match args.get(1) {
         Some(dir_name) => {
             if let Err(err) = fs::create_dir(dir_name) {
@@ -142,7 +142,7 @@ pub fn pwd() {
     }
 }
 
-pub fn read(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
+pub fn read(args: &[String], variables: &mut BTreeMap<String, String>) {
     for i in 1..args.len() {
         if let Some(arg_original) = args.get(i) {
             let arg = arg_original.trim();
@@ -158,7 +158,7 @@ pub fn read(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
     }
 }
 
-pub fn rm(args: &Vec<String>) {
+pub fn rm(args: &[String]) {
     match args.get(1) {
         Some(path) => {
             if fs::remove_file(path).is_err() {
@@ -169,7 +169,7 @@ pub fn rm(args: &Vec<String>) {
     }
 }
 
-pub fn rmdir(args: &Vec<String>) {
+pub fn rmdir(args: &[String]) {
     match args.get(1) {
         Some(path) => {
             if fs::remove_dir(path).is_err() {
@@ -180,7 +180,7 @@ pub fn rmdir(args: &Vec<String>) {
     }
 }
 
-pub fn run(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
+pub fn run(args: &[String], variables: &mut BTreeMap<String, String>) {
     let path = "/apps/shell/main.bin";
 
     let mut command = process::Command::new(path);
@@ -207,12 +207,12 @@ pub fn run(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
     }
 }
 
-pub fn sleep(args: &Vec<String>) {
+pub fn sleep(args: &[String]) {
     let secs = args.get(1).map_or(0, |arg| arg.to_num());
     thread::sleep_ms(secs as u32 * 1000);
 }
 
-pub fn touch(args: &Vec<String>) {
+pub fn touch(args: &[String]) {
     match args.get(1) {
         Some(file_name) => {
             if let Err(err) = File::create(file_name) {
--- a/src/builtin.rs
+++ b/src/builtin.rs
@@ -1,0 +1,239 @@
+use std::collections::BTreeMap;
+use std::string::String;
+use std::vec::Vec;
+use std::fs::{self, File};
+use std::io::{stdout, Read, Write};
+use std::env;
+use std::process;
+use std::thread;
+
+use super::to_num::ToNum;
+use super::set_var;
+use super::input_editor::readln;
+
+pub fn cat(args: &Vec<String>) {
+    let path = args.get(1).map_or(String::new(), |arg| arg.clone());
+
+    match File::open(&path) {
+        Ok(mut file) => {
+            let mut string = String::new();
+            match file.read_to_string(&mut string) {
+                Ok(_) => println!("{}", string),
+                Err(err) => {
+                    println!("Failed to read: {}: {}", path, err)
+                }
+            }
+        }
+        Err(err) => println!("Failed to open file: {}: {}", path, err),
+    }
+}
+
+pub fn cd(args: &Vec<String>) {
+    match args.get(1) {
+        Some(path) => {
+            if let Err(err) = env::set_current_dir(&path) {
+                println!("Failed to set current dir to {}: {}",
+                         path,
+                         err);
+            }
+        }
+        None => println!("No path given"),
+    }
+}
+
+pub fn echo(args: &Vec<String>) {
+    let echo = args.iter()
+        .skip(1)
+        .fold(String::new(),
+        |string, arg| string + " " + arg);
+    println!("{}", echo.trim());
+}
+
+pub fn free() {
+    match File::open("memory:") {
+        Ok(mut file) => {
+            let mut string = String::new();
+            match file.read_to_string(&mut string) {
+                Ok(_) => println!("{}", string),
+                Err(err) => println!("Failed to read: memory: {}", err),
+            }
+        }
+        Err(err) => println!("Failed to open file: memory: {}", err),
+    }
+}
+
+pub fn ls(args: &Vec<String>) {
+    let path = args.get(1).map_or(".".to_string(), |arg| arg.clone());
+
+    let mut entries = Vec::new();
+    match fs::read_dir(&path) {
+        Ok(dir) => {
+            for entry_result in dir {
+                match entry_result {
+                    Ok(entry) => {
+                        let directory = match entry.file_type() {
+                            Ok(file_type) => file_type.is_dir(),
+                            Err(err) => {
+                                println!("Failed to read file type: {}", err);
+                                false
+                            }
+                        };
+
+                        match entry.file_name().to_str() {
+                            Some(path_str) => {
+                                if directory {
+                                    entries.push(path_str.to_string() + "/")
+                                } else {
+                                    entries.push(path_str.to_string())
+                                }
+                            }
+                            None => {
+                                println!("Failed to convert path to string")
+                            }
+                        }
+                    }
+                    Err(err) => {
+                        println!("Failed to read entry: {}", err)
+                    }
+                }
+            }
+        }
+        Err(err) => {
+            println!("Failed to open directory: {}: {}", path, err)
+        }
+    }
+
+    entries.sort();
+
+    for entry in entries {
+        println!("{}", entry);
+    }
+}
+
+pub fn mkdir(args: &Vec<String>) {
+    match args.get(1) {
+        Some(dir_name) => {
+            if let Err(err) = fs::create_dir(dir_name) {
+                println!("Failed to create: {}: {}", dir_name, err);
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
+
+pub fn poweroff() {
+    match File::create("acpi:off") {
+        Err(err) => println!("Failed to remove power (error: {})", err),
+        Ok(_) => println!("I see dead people"),
+    }
+}
+
+pub fn ps() {
+    match File::open("context:") {
+        Ok(mut file) => {
+            let mut string = String::new();
+            match file.read_to_string(&mut string) {
+                Ok(_) => println!("{}", string),
+                Err(err) => {
+                    println!("Failed to read: context: {}", err)
+                }
+            }
+        }
+        Err(err) => println!("Failed to open file: context: {}", err),
+    }
+}
+
+pub fn pwd() {
+    match env::current_dir() {
+        Ok(path) => {
+            match path.to_str() {
+                Some(path_str) => println!("{}", path_str),
+                None => println!("?"),
+            }
+        }
+        Err(err) => println!("Failed to get current dir: {}", err),
+    }
+}
+
+pub fn read(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
+    for i in 1..args.len() {
+        if let Some(arg_original) = args.get(i) {
+            let arg = arg_original.trim();
+            print!("{}=", arg);
+            if let Err(message) = stdout().flush() {
+                println!("{}: Failed to flush stdout", message);
+            }
+            if let Some(value_original) = readln() {
+                let value = value_original.trim();
+                set_var(variables, arg, value);
+            }
+        }
+    }
+}
+
+pub fn rm(args: &Vec<String>) {
+    match args.get(1) {
+        Some(path) => {
+            if fs::remove_file(path).is_err() {
+                println!("Failed to remove: {}", path);
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
+
+pub fn rmdir(args: &Vec<String>) {
+    match args.get(1) {
+        Some(path) => {
+            if fs::remove_dir(path).is_err() {
+                println!("Failed to remove: {}", path);
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
+
+pub fn run(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
+    let path = "/apps/shell/main.bin";
+
+    let mut command = process::Command::new(path);
+    for i in 1..args.len() {
+        if let Some(arg) = args.get(i) {
+            command.arg(arg);
+        }
+    }
+
+    match command.spawn() {
+        Ok(mut child) => {
+            match child.wait() {
+                Ok(status) => {
+                    if let Some(code) = status.code() {
+                        set_var(variables, "?", &format!("{}", code));
+                    } else {
+                        println!("{}: No child exit code", path);
+                    }
+                }
+                Err(err) => {
+                    println!("{}: Failed to wait: {}", path, err)
+                }
+            }
+        }
+        Err(err) => println!("{}: Failed to execute: {}", path, err),
+    }
+}
+
+pub fn sleep(args: &Vec<String>) {
+    let secs = args.get(1).map_or(0, |arg| arg.to_num());
+    thread::sleep_ms(secs as u32 * 1000);
+}
+
+pub fn touch(args: &Vec<String>) {
+    match args.get(1) {
+        Some(file_name) => {
+            if let Err(err) = File::create(file_name) {
+                println!("Failed to create: {}: {}", file_name, err);
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
--- a/src/builtin.rs
+++ b/src/builtin.rs
@@ -1,4 +1,4 @@
-use std::collections::BTreeMap;
+use std::collections::HashMap;
 use std::fs::{self, File};
 use std::io::{stdout, Read, Write};
 use std::env;
@@ -36,9 +36,7 @@ pub fn cd(args: &[String]) {
 }
 
 pub fn echo(args: &[String]) {
-    let echo = args.iter()
-                   .skip(1)
-                   .fold(String::new(), |string, arg| string + " " + arg);
+    let echo = args.join(" ");
     println!("{}", echo.trim());
 }
 
@@ -109,9 +107,8 @@ pub fn mkdir(args: &[String]) {
 }
 
 pub fn poweroff() {
-    match File::create("acpi:off") {
-        Err(err) => println!("Failed to remove power (error: {})", err),
-        Ok(_) => println!("I see dead people"),
+    if let Err(err) = File::create("acpi:off") {
+        println!("Failed to remove power (error: {})", err);
     }
 }
 
@@ -140,12 +137,13 @@ pub fn pwd() {
     }
 }
 
-pub fn read(args: &[String], variables: &mut BTreeMap<String, String>) {
+pub fn read(args: &[String], variables: &mut HashMap<String, String>) {
+    let mut out = stdout();
     for i in 1..args.len() {
         if let Some(arg_original) = args.get(i) {
             let arg = arg_original.trim();
             print!("{}=", arg);
-            if let Err(message) = stdout().flush() {
+            if let Err(message) = out.flush() {
                 println!("{}: Failed to flush stdout", message);
             }
             if let Some(value_original) = readln() {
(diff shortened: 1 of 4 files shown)


src/main.rs
```diff
@@ -36,7 +36,14 @@ pub mod expansion;
 pub struct Command {
     pub name: &'static str,
     pub help: &'static str,
-    pub main: Box<Fn(&Vec<String>, &mut BTreeMap<String, String>, &mut Vec<Mode>)>,
+    pub main: Box<Fn(&Vec<String>, &mut Shell)>,
```
Ticki: I would use a slice here instead of Vec.

src/main.rs
```diff
@@ -36,7 +36,14 @@ pub mod expansion;
 pub struct Command {
     pub name: &'static str,
     pub help: &'static str,
```
Ticki: You might want to make this a function, since it's not necessarily being static.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+use std::io::{stdout, Read, Write};
+use std::env;
+use std::process;
+use std::thread;
+
+use super::to_num::ToNum;
+use super::set_var;
+use super::input_editor::readln;
+
+pub fn cat(args: &Vec<String>) {
```
Ticki: Cat shouldn't be a built-in.

Ticki: Again, use &[String] instead.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    let path = args.get(1).map_or(String::new(), |arg| arg.clone());
+
+    match File::open(&path) {
+        Ok(mut file) => {
```
Ticki: You should fold this match.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+            match file.read_to_string(&mut string) {
+                Ok(_) => println!("{}", string),
+                Err(err) => println!("Failed to read: {}: {}", path, err),
+            }
+        }
+        Err(err) => println!("Failed to open file: {}: {}", path, err),
+    }
+}
+
+pub fn cd(args: &Vec<String>) {
```
Ticki: Vec -> Slice.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    match args.get(1) {
+        Some(path) => {
+            if let Err(err) = env::set_current_dir(&path) {
```
Ticki: This if can be folded into the `Some(path)` arm.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+                println!("Failed to set current dir to {}: {}", path, err);
+            }
+        }
+        None => println!("No path given"),
+    }
+}
+
+pub fn echo(args: &Vec<String>) {
```
Ticki: Vec -> Slice

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    let echo = args.iter()
```
Ticki: Use `join` from slice::SliceConcatExt instead.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+                   .skip(1)
+                   .fold(String::new(), |string, arg| string + " " + arg);
```
Ticki: Unecessary clone.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    println!("{}", echo.trim());
+}
+
+pub fn free() {
+    match File::open("memory:") {
+        Ok(mut file) => {
+            let mut string = String::new();
+            match file.read_to_string(&mut string) {
```
Ticki: Can be folded.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+                Ok(_) => println!("{}", string),
+                Err(err) => println!("Failed to read: memory: {}", err),
+            }
+        }
+        Err(err) => println!("Failed to open file: memory: {}", err),
+    }
+}
+
+pub fn ls(args: &Vec<String>) {
```
Ticki: Vec -> Slice

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    let path = args.get(1).map_or(".".to_string(), |arg| arg.clone());
+
+    let mut entries = Vec::new();
```
Ticki: The following code can easily be simplified:

src/builtin.rs
Ticki: Why built-in?

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    }
+
+    entries.sort();
+
+    for entry in entries {
+        println!("{}", entry);
+    }
+}
+
+pub fn mkdir(args: &Vec<String>) {
```
Ticki: Why is mkdir built-in?

Ticki: Take slice as arg.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+        Some(dir_name) => {
+            if let Err(err) = fs::create_dir(dir_name) {
+                println!("Failed to create: {}: {}", dir_name, err);
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
+
+pub fn poweroff() {
```
Ticki: Shouldn't be built-in either.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    match File::create("acpi:off") {
+        Err(err) => println!("Failed to remove power (error: {})", err),
+        Ok(_) => println!("I see dead people"),
+    }
+}
+
+pub fn ps() {
```
Ticki: Again, this should be moved to core utils.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
+
+pub fn poweroff() {
+    match File::create("acpi:off") {
+        Err(err) => println!("Failed to remove power (error: {})", err),
+        Ok(_) => println!("I see dead people"),
```
Ticki: wtf?

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+            match path.to_str() {
+                Some(path_str) => println!("{}", path_str),
+                None => println!("?"),
+            }
+        }
+        Err(err) => println!("Failed to get current dir: {}", err),
+    }
+}
+
+pub fn read(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
```
Ticki: Should be moved to core-utils.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    for i in 1..args.len() {
+        if let Some(arg_original) = args.get(i) {
+            let arg = arg_original.trim();
+            print!("{}=", arg);
+            if let Err(message) = stdout().flush() {
```
Ticki: Don't keep calling `stdout()`. Just call it in the start and use that variable.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+            }
+            if let Some(value_original) = readln() {
+                let value = value_original.trim();
+                set_var(variables, arg, value);
+            }
+        }
+    }
+}
+
+pub fn rm(args: &Vec<String>) {
```
Ticki: Should be moved to core-utils.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+        Some(path) => {
+            if fs::remove_file(path).is_err() {
+                println!("Failed to remove: {}", path);
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
+
+pub fn rmdir(args: &Vec<String>) {
```
Ticki: This too.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+        Some(path) => {
+            if fs::remove_dir(path).is_err() {
+                println!("Failed to remove: {}", path);
+            }
+        }
+        None => println!("No name provided"),
+    }
+}
+
+pub fn run(args: &Vec<String>, variables: &mut BTreeMap<String, String>) {
```
Ticki: This too.

Ticki: Also, take a slice instead.

src/builtin.rs
Ticki: Also, take a slice instead.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+                    }
+                }
+                Err(err) => println!("{}: Failed to wait: {}", path, err),
+            }
+        }
+        Err(err) => println!("{}: Failed to execute: {}", path, err),
+    }
+}
+
+pub fn sleep(args: &Vec<String>) {
```
Ticki: This too. Also, take a slice instead.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+    let secs = args.get(1).map_or(0, |arg| arg.to_num());
+    thread::sleep_ms(secs as u32 * 1000);
+}
+
+pub fn touch(args: &Vec<String>) {
+    match args.get(1) {
+        Some(file_name) => {
+            if let Err(err) = File::create(file_name) {
```
Ticki: This will clear the file, which is not intented.

src/main.rs
```diff
@@ -36,7 +36,14 @@ pub mod expansion;
...
-    pub main: Box<Fn(&Vec<String>, &mut BTreeMap<String, String>, &mut Vec<Mode>)>,
+    pub main: Box<Fn(&Vec<String>, &mut Shell)>,
+}
+
+/// This struct will contain all of the data structures related to this
+/// instance of the shell.
+pub struct Shell<'a> {
+    pub variables: &'a mut BTreeMap<String, String>,
```
Ticki: Prefer HashMap, BTreeMap can quickly get slow as the number of variables grow.

src/main.rs
```diff
@@ -36,7 +36,14 @@ pub mod expansion;
...
+    pub modes: &'a mut Vec<Mode>,
```
Ticki: Why don't you let `Shell` own the vec instead of borrowing it?

src/main.rs
```diff
@@ -48,23 +55,8 @@ impl Command {
...
-                                        match file.read_to_string(&mut string) {
-                                            Ok(_) => println!("{}", string),
-                                            Err(err) => {
-                                                println!("Failed to read: {}: {}", path, err)
-                                            }
-                                        }
-                                    }
-                                    Err(err) => println!("Failed to open file: {}: {}", path, err),
-                                }
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
```
Ticki: Use `box` keyword instead, this will construct the closure in-place.

src/main.rs
```diff
@@ -48,23 +55,8 @@ impl Command {
...
+                                builtin::cat(args);
```
Ticki: You can cast the method directly. No need to have a closure just calling `cat`.

src/main.rs
Ticki: You might not want to take a mutable reference to `Shell` here. Just let the closure capture it.

src/main.rs
```diff
@@ -72,19 +64,8 @@ impl Command {
...
-                                        if let Err(err) = env::set_current_dir(&path) {
-                                            println!("Failed to set current dir to {}: {}",
-                                                     path,
-                                                     err);
-                                        }
-                                    }
-                                    None => println!("No path given"),
-                                }
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::cd(args);
```
Ticki: Again, just use direct casting.

src/main.rs
```diff
@@ -92,14 +73,8 @@ impl Command {
...
-                            main: Box::new(|args: &Vec<String>,
-                                            _: &mut BTreeMap<String, String>,
-                                            _: &mut Vec<Mode>| {
-                                let echo = args.iter()
-                                               .skip(1)
-                                               .fold(String::new(),
-                                                     |string, arg| string + " " + arg);
-                                println!("{}", echo.trim());
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::echo(args);
```
Ticki: And here too.

src/main.rs
```diff
@@ -107,28 +82,15 @@ impl Command {
                         Command {
                             name: "exit",
                             help: "To exit the curent session",
-                            main: Box::new(|_: &Vec<String>,
-                                            _: &mut BTreeMap<String, String>,
-                                            _: &mut Vec<Mode>| {}),
+                            main: Box::new(|_: &Vec<String>, _: &mut Shell| {}),
```
Ticki: This does not do anything.

src/main.rs
```diff
@@ -107,28 +82,15 @@ impl Command {
...
-                                        let mut string = String::new();
-                                        match file.read_to_string(&mut string) {
-                                            Ok(_) => println!("{}", string),
-                                            Err(err) => println!("Failed to read: memory: {}", err),
-                                        }
-                                    }
-                                    Err(err) => println!("Failed to open file: memory: {}", err),
-                                }
+                            main: Box::new(|_: &Vec<String>, _: &mut Shell| {
+                                builtin::free();
```
Ticki: Casting is useful here too.

src/main.rs
```diff
@@ -136,58 +98,8 @@ impl Command {
...
-                                    }
-                                }
-
-                                entries.sort();
-
-                                for entry in entries {
-                                    println!("{}", entry);
-                                }
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::ls(args);
```
Ticki: Here too.

src/main.rs
```diff
@@ -196,17 +108,8 @@ impl Command {
...
-                                match args.get(1) {
-                                    Some(dir_name) => {
-                                        if let Err(err) = fs::create_dir(dir_name) {
-                                            println!("Failed to create: {}: {}", dir_name, err);
-                                        }
-                                    }
-                                    None => println!("No name provided"),
-                                }
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::mkdir(args);
```
Ticki: And here, too.

src/main.rs
```diff
@@ -229,21 +127,8 @@ impl Command {
...
-                                            Ok(_) => println!("{}", string),
-                                            Err(err) => {
-                                                println!("Failed to read: context: {}", err)
-                                            }
-                                        }
-                                    }
-                                    Err(err) => println!("Failed to open file: context: {}", err),
-                                }
+                            main: Box::new(|_: &Vec<String>, _: &mut Shell| {
+                                builtin::ps();
```
Ticki: And here, too.

src/main.rs
```diff
@@ -251,18 +136,8 @@ impl Command {
...
-                                    Ok(path) => {
-                                        match path.to_str() {
-                                            Some(path_str) => println!("{}", path_str),
-                                            None => println!("?"),
-                                        }
-                                    }
-                                    Err(err) => println!("Failed to get current dir: {}", err),
-                                }
+                            main: Box::new(|_: &Vec<String>, _: &mut Shell| {
+                                builtin::pwd();
```
Ticki: And here, too.

src/main.rs
```diff
@@ -293,17 +154,8 @@ impl Command {
...
-                                match args.get(1) {
-                                    Some(path) => {
-                                        if fs::remove_file(path).is_err() {
-                                            println!("Failed to remove: {}", path);
-                                        }
-                                    }
-                                    None => println!("No name provided"),
-                                }
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::rm(args);
```
Ticki: And here, too.

src/main.rs
```diff
@@ -311,17 +163,8 @@ impl Command {
...
-                                match args.get(1) {
-                                    Some(path) => {
-                                        if fs::remove_dir(path).is_err() {
-                                            println!("Failed to remove: {}", path);
-                                        }
-                                    }
-                                    None => println!("No name provided"),
-                                }
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::rmdir(args);
```
Ticki: And here, too.

src/main.rs
```diff
@@ -366,11 +182,8 @@ impl Command {
                             name: "sleep",
                             help: "Make a sleep in the current session\n    sleep \
                                    <number_of_seconds>",
-                            main: Box::new(|args: &Vec<String>,
-                                            _: &mut BTreeMap<String, String>,
-                                            _: &mut Vec<Mode>| {
-                                let secs = args.get(1).map_or(0, |arg| arg.to_num());
-                                thread::sleep_ms(secs as u32 * 1000);
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::sleep(args);
```
Ticki: And here, too.

src/main.rs
```diff
@@ -381,17 +194,8 @@ impl Command {
...
-                                match args.get(1) {
-                                    Some(file_name) => {
-                                        if let Err(err) = File::create(file_name) {
-                                            println!("Failed to create: {}: {}", file_name, err);
-                                        }
-                                    }
-                                    None => println!("No name provided"),
-                                }
+                            main: Box::new(|args: &Vec<String>, _: &mut Shell| {
+                                builtin::touch(args);
```
Ticki: And here, too.

src/main.rs
```diff
@@ -438,19 +240,16 @@ pub struct Mode {
...
 
-fn on_command(command_string: &str,
-              commands: &BTreeMap<String, Command>,
-              variables: &mut BTreeMap<String, String>,
-              modes: &mut Vec<Mode>) {
+fn on_command(command_string: &str, commands: &BTreeMap<String, Command>, shell: &mut Shell) {
     // Show variables
     if command_string == "$" {
-        for (key, value) in variables.iter() {
+        for (key, value) in shell.variables.iter() {
```
Ticki: You might want to keep `variables` sorted.

src/main.rs
```diff
@@ -438,19 +240,16 @@ pub struct Mode {
     value: bool,
 }
 
-fn on_command(command_string: &str,
-              commands: &BTreeMap<String, Command>,
-              variables: &mut BTreeMap<String, String>,
-              modes: &mut Vec<Mode>) {
+fn on_command(command_string: &str, commands: &BTreeMap<String, Command>, shell: &mut Shell) {
```
Ticki: This should be a method defined on `Shell`.

Ticki: Also, use `HashMap` instead.

Ticki: Also, s/string/&str/

src/main.rs
```diff
@@ -438,19 +240,16 @@ pub struct Mode {
...
     if command_string == "$" {
-        for (key, value) in variables.iter() {
+        for (key, value) in shell.variables.iter() {
             println!("{}={}", key, value);
         }
         return;
     }
 
-    let mut tokens: Vec<Token> = expand_tokens(&mut tokenize(command_string), variables);
+    let mut tokens: Vec<Token> = expand_tokens(&mut tokenize(command_string), shell.variables);
```
Ticki: No need for annotation type here.

src/main.rs
```diff
@@ -438,19 +240,16 @@ pub struct Mode {
...
     let jobs: Vec<Job> = parse(&mut tokens);
```
Ticki: Neither here.

src/main.rs
```diff
@@ -486,13 +285,13 @@ fn on_command(command_string: &str,
...
 
-            modes.insert(0, Mode { value: value });
+            shell.modes.insert(0, Mode { value: value });
             continue;
         }
 
         if job.command == "else" {
             let mut syntax_error = false;
-            match modes.get_mut(0) {
+            match shell.modes.get_mut(0) {
```
Ticki: You might want to use `if let` instead.

src/main.rs
```diff
@@ -486,13 +285,13 @@ fn on_command(command_string: &str,
                 println!("No left hand side");
             }
 
-            modes.insert(0, Mode { value: value });
+            shell.modes.insert(0, Mode { value: value });
             continue;
         }
 
         if job.command == "else" {
```
Ticki: Avoid this type of comparations. Lookup is better.

src/main.rs
```diff
@@ -504,8 +303,8 @@ fn on_command(command_string: &str,
 
         if job.command == "fi" {
             let mut syntax_error = false;
-            if !modes.is_empty() {
-                modes.remove(0);
+            if !shell.modes.is_empty() {
+                shell.modes.remove(0);
```
Ticki: This can panic.

src/main.rs
```diff
@@ -590,8 +389,7 @@ fn print_prompt(modes: &Vec<Mode>) {
     }
 }
 
-fn run_external_commmand(args: Vec<String>,
-                         variables: &mut BTreeMap<String, String>) {
+fn run_external_commmand(args: Vec<String>, variables: &mut BTreeMap<String, String>) {
```
Ticki: s/BTreeMap/HashMap/

Ticki: Use a slice instead.

src/builtin.rs
skylerberg: What do you mean by "fold"? Do you mean use an `if-let`?

src/main.rs
```diff
@@ -604,29 +402,25 @@ fn run_external_commmand(args: Vec<String>,
                 match child.wait() {
                     Ok(status) => {
                         if let Some(code) = status.code() {
-                            set_var(variables,
-                                    "?",
-                                    &format!("{}", code));
+                            set_var(variables, "?", &format!("{}", code));
```
Ticki: `format!` have overhead. Why not just use to_string? That's cheaper.

src/main.rs
```diff
@@ -604,29 +402,25 @@ fn run_external_commmand(args: Vec<String>,
...
-                println!("{}: Failed to execute: {}", path, err)
-            }
+            Err(err) => println!("{}: Failed to execute: {}", path, err),
         }
     }
 }
 
 fn main() {
     let commands = Command::map();
-    let mut variables: BTreeMap<String, String> = BTreeMap::new();
```
Ticki: No need for type annotations here.

src/main.rs
```diff
@@ -604,29 +402,25 @@ fn run_external_commmand(args: Vec<String>,
...
-    let mut modes: Vec<Mode> = vec![];
+    let mut shell = Shell {
```
Ticki: You might want to make this into a method `new` instead.

src/builtin.rs
skylerberg: I am also not sure what you mean here.

src/builtin.rs
Ticki: I mean esentially merging the two matches into one.

Ticki:
Great :+1:

src/builtin.rs
Ticki: Use the `PATTERN if CONDITION =>` syntax.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
+use std::collections::BTreeMap;
+use std::string::String;
```
Ticki: No need for this. The prelude is working.

src/builtin.rs
```diff
@@ -0,0 +1,224 @@
...
+use std::vec::Vec;
```
Ticki: Same here.

src/main.rs
skylerberg: I have been using BTreeMap because HashMap is (or at least was not) implemented in libredox. Do you think I should change to HashMap?

src/builtin.rs
Ticki: Also, cat isn't a "print this file" command. It's for concatenation.

src/main.rs
Ticki: I implemented it a bunch of weeks ago. It should still be there.

skylerberg: I mean FromIterator was not implemented for HashMap, actually.

src/main.rs
skylerberg: This exists so that there is a help entry for exit. It should really abort the current process. I don't remember why, but there is a reason it is the way it is right now. I will open an issue to address this.

src/main.rs
Ticki: see https://github.com/redox-os/redox/pull/469

src/main.rs
skylerberg: I think I do want to do this because otherwise the shell is forever borrowed by the function that creates the closures. The closures last for pretty much the whole time the main process is running, so having the shell stuck there won't work.

src/main.rs
skylerberg: You mean make a help function instead of a help string? That could work, but I am not sure when it wouldn't be static.

src/builtin.rs
skylerberg: Agreed on all. I will leave it in for now, but I am looking forward to removing all the core-utils functions.

src/builtin.rs
skylerberg: Awesome! That is a lot cleaner.

src/builtin.rs
skylerberg: Since this should be moved to core-utils and is not implemented correctly here, I am just going to remove it.

src/main.rs
skylerberg: Thanks! I am going to switch to HashMap inside here now.

src/main.rs
skylerberg: Done. I also gave it ownership of variables.

src/main.rs
skylerberg: Cool. I did not know about this.

src/main.rs
skylerberg: It looks like the BTreeMap happens to keep it in alphabetical order. I will make sure that it sorts them when it prints them out when I change it to a hashmap.

src/main.rs
skylerberg: I am not sure what I should be looking up here.

src/main.rs
skylerberg: Can it still panic even though we know that the vec is not empty?

PR #26, opened 2016-01-17
funky grammar proposal

Ticki:


polymetric1:
Why is there a semicolon after the `let` statement? It makes sense but unnecessary.

Ticki:
If we want to avoid complications with the parser semicolons are a good idea.

polymetric1:
Will the parser execute input line by line?

Ticki:
> Will the parser execute input line by line?

No, that's not what a parser does. It parses it, execution is a whole other process.

polymetric1:
Alright I'm stupid. I have no experience with shell parsers so I should read up on it first

Ticki:
No problem. The process is something like this:

`Program -- Lexer (tokenizer) --> Token list -- Parser --> AST (abstract syntax tree) -- Preprocessor -- Executor --> Output`

PR #32, opened 2016-01-18
Lots of cleanup and refactor

stratact:
* Remove an unneeded clone
* Cleanup the way `#[derive]` is used on a few types
* Derive `Default` for `parser::Job` and use `Job::default()` instead of `Job::new()` which did the same thing
* Use BTreeMap for shell variables to have it automatically be sorted for reading
* Remove an unneeded binding in `on_command()`
* Use a type alias for `BTreeMap<String, String>` everywhere for easier maintenance

### DIFF
--- a/src/parser.rs
+++ b/src/parser.rs
@@ -1,25 +1,17 @@
 use super::tokenizer::Token;
 
-#[derive(Debug)]
-#[derive(PartialEq)]
-#[derive(Clone)]
+#[derive(Clone, Debug, Default, PartialEq)]
 pub struct Job {
     pub command: String,
     pub args: Vec<String>,
 }
 
-impl Job {
-    pub fn new() -> Job {
-        Job {
-            command: String::new(),
-            args: vec![],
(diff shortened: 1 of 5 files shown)


src/parser.rs
```diff
@@ -1,25 +1,17 @@
...
-        }
-    }
-}
+//impl Job {
+//}
 
 pub fn parse(tokens: &mut Vec<Token>) -> Vec<Job> {
     let mut jobs: Vec<Job> = vec![];
-    let mut job = Job::new();
+    let mut job = Job::default();
```
Ticki: You might want to use the `Default` trait here?

stratact: I already derived it for `Job` though. What's the concern here?

stratact: This is what happens: http://is.gd/T7Y1Gb

Ticki: Oh, right. It were a derive. My fault.

PR #36, opened 2016-01-24
Redo parser using `rust-peg`

skylerberg:
This adds a new parser that uses a parser generator library. Moving off of my custom parser is a huge step in the right direction.

This new implementation is really messy. My goal was to get to feature parity with the previous implementation and add enough tests to be able to refactor.

The next steps are to refactor the grammar, make ion not crash on parse errors, then start extending the grammar.


Ticki:
Cool!

src/expansion.rs
```diff
@@ -1,24 +1,26 @@
...
-                    } else {
-                        Token::Word(String::new())
-                    }
-                } else {
-                    Token::Word(word)
-                }
-            }
-            _ => token,
-        });
+pub fn expand_variables(mut jobs: Vec<Job>, variables: &Variables) -> Vec<Job> {
```
Ticki: Try to avoid taking `Vec`s as paramets, use slices instead.

src/peg.rs
```diff
@@ -0,0 +1,259 @@
...
+    / newline
+    / newline
+
+newline -> ()
+    = [\r\n]
+"#);
+
+
+#[cfg(test)]
+mod tests {
```
Ticki: Great, tests!

Cargo.toml
```diff
@@ -11,3 +11,6 @@ authors = [
   "Łukasz Niemier <lukasz@niemier.pl>",
   "Skyler Berg <skylertheberg@gmail.com>"
```
Ticki: You're the main contributor. Why not put your name at the top?

src/expansion.rs
stratact: I'll take care of this in my next PR. Thanks for point it out.

PR #37, opened 2016-01-24
Fearless Cleanup

stratact:


src/peg.rs
```diff
@@ -16,7 +16,7 @@ impl Job {
 }
 
 pub fn parse(code: &str) -> Vec<Job> {
-    job_list(code).unwrap()  // TODO don't unwrap, handle parse error
+    job_list(code).unwrap_or(vec![])
```
skylerberg: I didn't know about `unwrap_or`. Neat!

PR #38, opened 2016-01-24
Refactor `print_prompt()`

stratact:


src/main.rs
```diff
@@ -288,24 +288,13 @@ pub fn set_var(variables: &mut Variables, name: &str, value: &str) {
...
-            print!("+ ");
-        } else {
-            print!("- ");
-        }
-    }
+fn print_prompt(modes: &[Mode]) {
+    let mode = modes.iter().rev().fold(String::new(), |acc, mode| {
+        acc + if mode.value { "+ " } else { "- " }
+    });
+    print!("{}", mode.trim_right());
```
skylerberg: Why are you trimming right?

src/main.rs
```diff
@@ -288,24 +288,13 @@ pub fn set_var(variables: &mut Variables, name: &str, value: &str) {
...
-fn print_prompt(modes: &Vec<Mode>) {
-    for mode in modes.iter().rev() {
-        if mode.value {
-            print!("+ ");
-        } else {
-            print!("- ");
-        }
-    }
+fn print_prompt(modes: &[Mode]) {
+    let mode = modes.iter().rev().fold(String::new(), |acc, mode| {
```
skylerberg: Perhaps naming the string that we print out `mode` is confusing `mode` is also used in the fold. I am thinking something like `mode_prompt` or `prompt_prefix` would be more descriptive.

src/main.rs
stratact: because there is an extra space to the right?

src/main.rs
```diff
@@ -288,24 +288,13 @@ pub fn set_var(variables: &mut Variables, name: &str, value: &str) {
...
-    let cwd = match env::current_dir() {
-        Ok(path) => {
-            match path.to_str() {
-                Some(path_str) => path_str.to_string(),
-                None => "?".to_string(),
-            }
-        }
-        Err(_) => "?".to_string(),
-    };
+    let cwd = env::current_dir().ok().map_or("?".to_string(), |ref p| p.to_str().unwrap_or("?").to_string());
```
skylerberg: I like that this gets rid of the duplicate `to_string`s and the duplicate `"?"`. That said, this seems a little like code golf.

stratact: Well, it's pretty much idiomatic and shorter while being concise without the verbose match expressions.

src/main.rs
skylerberg: I think that was intentional. If you are in an else branch in an else branch it would say:
`- - user@ion#` or something like that.

stratact: Alright, will fix.

PR #45, opened 2016-01-28
Move rogue functions to methods and cleanup

stratact:
The problem with the way `main.rs` was written, was that it ignored a fundamental rule to encapsulation completely and when you do that, you end doing more work for the object rather than having the object do the work for you. Having the object do the work, allows the programmer to be code correct without having to think about it and not needing to mess with the object's data. So, I moved all of the rogue functions into `Shell` methods which is where they seem to belong, made all `Shell` fields private, cleaned up the signatures, and let the `Shell` object speak for itself.

As a result the code ends up looking cleaner this way and makes it easier for the programmer, API-wise.

### DIFF
--- a/src/expansion.rs
+++ b/src/expansion.rs
@@ -1,25 +1,32 @@
 use super::peg::Job;
-use super::Variables;
+use super::Shell;
 
-pub fn expand_variables(jobs: &mut [Job], variables: &Variables) {
-    for mut job in &mut jobs[..] {
-        job.command = expand_string(&job.command, variables).to_string();
-        job.args = job.args
-                      .iter()
-                      .map(|original: &String| expand_string(&original, variables).to_string())
-                      .collect();
-    }
+pub trait Expand {
+    fn expand_variables(&self, jobs: &mut [Job]);
+    fn expand_string<'a>(&'a self, original: &'a str) -> &'a str;
 }
 
-#[inline]
-fn expand_string<'a>(original: &'a str, variables: &'a Variables) -> &'a str {
-    if original.starts_with("$") {
-        if let Some(value) = variables.get(&original[1..]) {
-            &value
-        } else {
-            ""
+impl Expand for Shell {
+    fn expand_variables(&self, jobs: &mut [Job]) {
+        for mut job in &mut jobs[..] {
+            job.command = self.expand_string(&job.command).to_string();
+            job.args = job.args
+                .iter()
+                .map(|original: &String| self.expand_string(&original).to_string())
+                .collect();
         }
-    } else {
-        original
+    }
+
+    #[inline]
+    fn expand_string<'a>(&'a self, original: &'a str) -> &'a str {
+        if original.starts_with("$") {
+            if let Some(value) = self.variables.get(&original[1..]) {
+                &value
+            } else {
+                ""
+            }
+        } else {
+            original
+        }
     }
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -26,10 +26,10 @@ pub type Variables = BTreeMap<String, String>;
 /// This struct will contain all of the data structures related to this
 /// instance of the shell.
 pub struct Shell {
-    pub variables: Variables,
-    pub modes: Vec<Mode>,
-    pub directory_stack: DirectoryStack,
-    pub history: VecDeque<String>,
+    variables: Variables,
+    modes: Vec<Mode>,
+    directory_stack: DirectoryStack,
+    history: VecDeque<String>,
 }
 
 impl Shell {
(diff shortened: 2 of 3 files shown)


Ticki:
:+1:

src/expansion.rs
```diff
@@ -1,25 +1,32 @@
...
 
-pub fn expand_variables(jobs: &mut [Job], variables: &Variables) {
-    for mut job in &mut jobs[..] {
-        job.command = expand_string(&job.command, variables).to_string();
-        job.args = job.args
-                      .iter()
-                      .map(|original: &String| expand_string(&original, variables).to_string())
-                      .collect();
-    }
+pub trait Expand {
```
Ticki: Could you add some docs here?

Ticki: Why is this a trait? It should just be a method defined on `Shell`.

src/main.rs
```diff
@@ -21,15 +21,17 @@ pub mod input_editor;
...
+}
 
 /// This struct will contain all of the data structures related to this
 /// instance of the shell.
 pub struct Shell {
-    pub variables: Variables,
-    pub modes: Vec<Mode>,
-    pub directory_stack: DirectoryStack,
-    pub history: VecDeque<String>,
+    variables: BTreeMap<String, String>,
```
Ticki: Why BTreeMap, and not HashMap?

src/main.rs
```diff
@@ -21,15 +21,17 @@ pub mod input_editor;
 pub mod peg;
 pub mod expansion;
 
-pub type Variables = BTreeMap<String, String>;
+pub struct Mode {
```
Ticki: Might want to add some docs here.

src/main.rs
```diff
@@ -21,15 +21,17 @@ pub mod input_editor;
...
 
 /// This struct will contain all of the data structures related to this
 /// instance of the shell.
 pub struct Shell {
-    pub variables: Variables,
-    pub modes: Vec<Mode>,
-    pub directory_stack: DirectoryStack,
-    pub history: VecDeque<String>,
+    variables: BTreeMap<String, String>,
+    modes: Vec<Mode>,
```
Ticki: This isn't supposed to be mutated, so it might be inappropriate to use `Vec`.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
             history: VecDeque::new(),
         }
     }
+
+    pub fn print_prompt(&self) {
+        let prompt_prefix = self.modes.iter().rev().fold(String::new(), |acc, mode| {
+            acc +
```
Ticki: Line breaks here is bad style.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
             history: VecDeque::new(),
         }
     }
+
+    pub fn print_prompt(&self) {
+        let prompt_prefix = self.modes.iter().rev().fold(String::new(), |acc, mode| {
```
Ticki: Use `join()` from `SliceConcatExt` instead.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+                                                 |ref p| p.to_str().unwrap_or("?").to_string());
+
+        print!("ion:{}# ", cwd);
+        if let Err(message) = stdout().flush() {
+            println!("{}: failed to flush prompt to stdout", message);
+        }
+    }
+
+    fn on_command(&mut self, command_string: &str, commands: &HashMap<&str, Command>) {
+        let max_history: usize = 1000; // TODO temporary, make this configurable
```
Ticki: Just fetch this from `self.variables`

src/expansion.rs
stratact: I need to ask @skylerberg if he would be okay to include them into `Shell` or if wants to keep the expansion stuff separate. I just made it into a trait to keep it modular for now. As for documentation, I don't know what say.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+            self.history.pop_front();
+        }
+        self.history.push_back(command_string.to_string());
+
+        // Show variables
+        if command_string == "$" {
+            for (key, value) in self.variables.iter() {
+                println!("{}={}", key, value);
+            }
+            return;
```
Ticki: Use `else` instead.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+
+        let mut jobs = parse(command_string);
+        self.expand_variables(&mut jobs);
+
+        // Execute commands
+        for job in jobs.iter() {
+            if job.command == "if" {
+                let mut value = false;
+
+                if let Some(left) = job.args.get(0) {
```
Ticki: This is hacky. Proper parsing is better.

src/main.rs
stratact: Actually, it was you that made the suggestion to BTreeMap, so that vars should also be automatically sorted upon insert. :stuck_out_tongue_closed_eyes:

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+                if !self.modes.is_empty() {
+                    self.modes.remove(0);
+                } else {
+                    println!("Syntax error: fi found with no previous if");
+                }
+                continue;
+            }
+
+            let mut skipped: bool = false;
+            for mode in self.modes.iter() {
```
Ticki: No need for `iter()` here.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+            if job.command == "fi" {
+                if !self.modes.is_empty() {
+                    self.modes.remove(0);
+                } else {
+                    println!("Syntax error: fi found with no previous if");
+                }
+                continue;
+            }
+
+            let mut skipped: bool = false;
```
Ticki: No need for annotating the type.

src/main.rs
stratact: Need to get info from @skylerberg

src/main.rs
Ticki: Use `any()` instead and assign it directly to `skipped`.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+                    break;
+                }
+            }
+            if skipped {
+                continue;
+            }
+
+            // Set variables
+            if let Some(i) = job.command.find('=') {
+                let name = job.command[0..i].trim();
```
Ticki: `..i` is sufficient.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+
+                self.set_var(name, &value);
+                continue;
+            }
+
+            // Commands
+            let mut args = job.args.clone();
+            args.insert(0, job.command.clone());
+            if let Some(command) = commands.get(&job.command.as_str()) {
+                (*command.main)(&args, self);
```
Ticki: No need for deref here.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+            } else {
+                self.run_external_commmand(args);
+            }
+        }
+    }
+
+    fn run_external_commmand(&mut self, args: Vec<String>) {
+        if let Some(path) = args.get(0) {
+            let mut command = process::Command::new(path);
+            for i in 1..args.len() {
```
Ticki: Note that `1` is not iterated over.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+                if let Some(arg) = args.get(i) {
+                    command.arg(arg);
+                }
+            }
+            match command.spawn() {
+                Ok(mut child) => {
+                    match child.wait() {
```
Ticki: Use `if let` here instead.

src/main.rs
stratact: I think this was caused by use of rustfmt, which I didn't do.

src/main.rs
```diff
@@ -42,6 +44,174 @@ impl Shell {
...
+                }
+                Err(err) => println!("{}: Failed to execute: {}", path, err),
+            }
+        }
+    }
+
+    pub fn set_var(&mut self, name: &str, value: &str) {
+        if name.is_empty() {
+            return;
+        }
```
Ticki: `else` is your friend!

src/main.rs
Ticki: rekt

Ticki:
pinging @skylerberg, since many of these comments are orginated in code of him.

Ticki:
Nice stuff tho!

src/main.rs
stratact: I think they want to get the value from a config file rather than a shell var.

src/main.rs
stratact: doesn't `return;` here explicitly exit the function though?

src/main.rs
Ticki: Well, why not fetch it from a variable, and then make it configurable by setting that var in the rc file?

src/main.rs
stratact: Didn't know about this one. Love combinators.

src/main.rs
Ticki: Yeah, it's correct, but it is not idiomatic, since you can easily fold this into the `if` statement.

src/main.rs
Ticki: It's awesome.

src/main.rs
stratact: Rustc didn't like me removing the deref here

Ticki: Oh. I thought it auto-deref'd function calls. Well, TIL.

src/main.rs
stratact: so just `else;` is fine here?

Ticki: Well, you can just add an `else` clause and wrap the rest of the function it it, deleting the `return;` statement.

stratact: fixed.

src/main.rs
stratact: I'm not sure if using `if let` matters here, since there is also matching for `Err(e)` as well?

skylerberg:
This looks like a really awesome change set! I am looking forward to going over this in about 5 hours.

stratact:
Bon appetit!

src/main.rs
skylerberg: I agree with Ticki that this should be a variable and that it should be settable rc-like file. That is not really in the scope of this change, so I will open an issue for it.

src/main.rs
skylerberg: If we are matching an `err` and we use `if let` how would we get the error message?

src/expansion.rs
```diff
@@ -1,25 +1,32 @@
...
+    fn expand_variables(&self, jobs: &mut [Job]);
+    fn expand_string<'a>(&'a self, original: &'a str) -> &'a str;
```
skylerberg: Does this have the unintended consequence of making `expand_string` public? It is meant to just be a helper method and should not be seen outside this module.

src/expansion.rs
skylerberg: I appreciate the intent here to turn functions into methods, but I don't think it makes sense in this case. There is no need for the expansion logic to know that `Shell` even exists. There are two things that I think would be reasonable here. 1) Leave it how it was before and have it take a Map of some sort. 2) Create a `Variables` struct. Since option two has other ramifications throughout the code I will add my thoughts as a comment on the PR rather than an inline comment in the file.

src/main.rs
skylerberg: Do you mean that `0` is not iterated over? If so, that is intended.

src/main.rs
skylerberg: Docs would be nice, but the whole `modes` thing is going to get a massive rework eventually, so I am not overly concerned.

skylerberg:
We may want to add `Variables` struct that wraps a Map of some sort. The advantage of this is that it is a natural place to put `expand_variables`, `read`, and `set_var`. I prefer this to the current solution because I want my methods to know as little as possible about the rest of the shell. For example, `read` does not need access to the entire shell. `expand_variables` is an interaction involving a string and a set of variables, not a shell and a string. The disadvantage is having a struct that has just a single member (the map) inside of it.

This can be taken care of in a future patch. I am going to see if anyone has anything else to add on chat and then merge this.

PR #48, opened 2016-01-29
Make travis check for formatting.

skylerberg:


skylerberg:
Not only does this not work right now, but I also appear to have duplicated commits from stratact. I guess I will have to level up my git-fu once it comes time to actually merge this.

PR #52, opened 2016-01-29
Implement exit

polymetric1:


skylerberg:
This change should get rid of the logic that currently handles the user typing exit in the main loop. Also, I thought there was a reason that I left things the way they were. I think it had to do with being compatible with redox. For that reason, I am going to hold off on merging this until Ion works on redox again and I can test this out.

jackpot51:
This change is ok now, I recently added process::exit to std

skylerberg:
@jackpot51, That must have been the compatibility issue I was thinking of. Excellent!

PR #53, opened 2016-01-30
Code cleanup

stratact:


skylerberg:
See CI log for why this does not compile.

stratact:
Fixing it!

stratact:
Should be mergeable now.

PR #54, opened 2016-01-30
Make set_var private and use let where it was used

skylerberg:
There was no good reason to expose set_var and have two ways to do the
same thing.


skylerberg:
This is actually a bad idea because `let_` takes it's args as a view that expects the first one to be `let`. In light of this, we should never use builtins for internal logic.

PR #56, opened 2016-01-30
remove exit "hack" before on_command()

polymetric1:


skylerberg:
It makes me happy every time someone can remove something like this :)

PR #59, opened 2016-01-30
Make cd use dirstack and limit dirstack size

skylerberg:


src/directory_stack.rs
```diff
@@ -1,14 +1,21 @@
+use std::collections::VecDeque;
 use std::env::{set_current_dir, current_dir};
 use std::path::PathBuf;
 
 pub struct DirectoryStack {
-    dirs: Vec<PathBuf>, // The top is always the current directory
+    dirs: VecDeque<PathBuf>, // The top is always the current directory
```
Ticki: May I ask: Why don't you use a ring buffer?

PR #61, opened 2016-02-06
Try to fix the usage of Strings

skylerberg:
The results are somewhat mixed, but I think this is an improvement.


src/main.rs
```diff
@@ -135,17 +135,15 @@ impl Shell {
             }
 
             // Commands
-            let mut args = job.args.clone();
```
skylerberg: Cleaning this up is the best part of this change.

PR #73, opened 2016-02-08
Fix `popd` bug, make bare `cd` take you home, add `cd -`

skylerberg:


src/directory_stack.rs
```diff
@@ -24,53 +24,86 @@ impl DirectoryStack {
...
-            return;
+            println!("Failed to get home directory");
+        }
+    }
+
+    fn switch_to_previous_directory(&mut self) {
+        if let Some(prev) = self.get_previous_dir()
+                                .map(|path| path.to_string_lossy().to_string()) {
+            println!("{}", prev);
+            self.change_and_push_dir(&prev);
```
Ticki: Actually, this is not the POSIX behavior. What POSIX does is to swap the last and next to last directories in the dirstack. I believe this is a better behavior, since you can already `popd`, anyways. POSIX's `cd -` serves a different purpose than `popd` (both are useful!), and I don't think duplicate functionality is really necessary in this case.

jackpot51:
I agree with ticki, swapping the top of the stack with the current directory is what I expect `cd -` to do

skylerberg:
I will switch it because I have no preference either way, but I do have a couple technical notes.

First, there is no POSIX specification for directory stacks as far as I know. `sh` does not have `pushd` and friends. Second, in the shells that do have directory stacks, the only one I know where `cd` manipulates it is zsh. IMO zsh does the directory stack wrong because adding a directory that previously was in the stack actually brings it to the front rather than adding it again. This is wrong because if you have an algorithm that matches pushes and pops it is possible for zsh to error if you pushed the same dir twice. Given that, even if `cd -` did push rather than swap in `zsh`, it would actually appear as if it was a swap because it would bring it to the front. So to my knowledge no other shell has made a decision regarding this so there is no precedent.

But, yeah, defining `cd -` as swapping the top two elements of the directory stack seems pretty reasonable to me.

Ticki:
Oh, right. I meant to say Zsh, not POSIX.

Ticki:
Either way, it is reasonable to have `cd -` to swap `PWD` and `DIR~`, since you can pop simply using `popd`.

skylerberg:
Again, not disagreeing and I will change this, but the way I have it is not the same as `popd`. `popd` takes you to the previous directory by removing an element. My current implementation takes you to the previous directory by putting it on the stack again. So `popd` decreases the size of the stack, with swapping it will not change the size of the stack, and the way I have it now increases the size of the stack.

PR #95, opened 2016-03-01
Fix the let command

nounoursheureux:
The `let` command now works like this:
- `let x=foo` creates a variable called `x` with a value of `foo`
- `let x` deletes the variable `x`
- `let` lists all the variables with their value

### DIFF
--- a/src/variables.rs
+++ b/src/variables.rs
@@ -100,7 +100,7 @@ mod tests {
     #[test]
     fn let_and_expand_a_variable() {
         let mut variables = Variables::new();
-        variables.let_(vec!["FOO", "BAR"]);
+        variables.let_(vec!["let", "FOO", "=", "BAR"]);
         let expanded = variables.expand_string("$FOO");
         assert_eq!("BAR", expanded);
     }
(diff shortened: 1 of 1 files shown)


nounoursheureux:
Oops, I broke the tests. I'm gonna fix this.

skylerberg:
Cool, that looks better @nounoursheureux. In the future it would be nice to have each of the following work as well:
`let x =y`
`let x= y`
`let x = y`
and maybe `let x y`.
We probably need a documentation change somewhere for this change to `let`. If there isn't somewhere in the documentation/code that explains how `let` works, then there probably should be.

Ticki:
I would rather avoid `let x y`, because having one uniform way to do it is always preferable.

nounoursheureux:
> Cool, that looks better @nounoursheureux. In the future it would be nice to have each of the following work as well:
let x =y
let x= y
let x = y

It works, I can add tests for this if you want.

>  We probably need a documentation change somewhere for this change to let. If there isn't somewhere in the documentation/code that explains how let works, then there probably should be.

Agreed. I will update SYNTAX.md.

PR #96, opened 2016-03-01
Less Obnoxious Error

mmstick:
Instead of the current verbose error message when an invalid command is entered, what if we were to print a traditional error message like so:

ion:/home/mmstick/Programming/Rust/ion# HELP
ion: command not found: HELP


skylerberg:
Thanks for improving this. Now that `err` isn't used the match can be changed to an if-let as well.

PR #100, opened 2016-03-02
Remove curly braces from purposed syntax

skylerberg:
I didn't update the proposed grammar. Since it is rough and incomplete, I didn't want to take the time to maintain it. If I or someone else does an overhaul at some point, that would be great.

### DIFF
--- a/SYNTAX.md
+++ b/SYNTAX.md
@@ -24,7 +24,7 @@
 
 ## Proposed Syntax
 
-A LR(k) grammar. Something like this (pseudocode):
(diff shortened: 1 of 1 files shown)


SYNTAX.md
```diff
@@ -24,7 +24,7 @@
 
 ## Proposed Syntax
 
-A LR(k) grammar. Something like this (pseudocode):
+A LR(k) grammar. This is a rough brainstorm and is somewhat out of sync with the exampels:
```
jackpot51: examples

skylerberg: Thanks.

PR #103, opened 2016-03-03
Exit With Previous Status

mmstick:
This will allow Ion to exit with the last command's exit status. The previous command's exit status is stored inside the `History` structure attached to the `Shell`.


jackpot51:
It must also process::exit with that code at the end of main()

mmstick:
It should be good now.

Ticki:
Nice work, @mmstick!

Issue #70, opened 2016-02-07
Support user defined functions in ion


skylerberg:
I think we should do functions differently than other shells. I think they should be able to have a set number of arguments like a normal scripting language.

The first step toward this that I want to implement is zero argument functions. These will basically be like aliases. Like Fish, I don't think ion needs aliases when functions can do the job.

nounoursheureux:
We should define the syntax for the functions. We could do something like this:
```
<function_keyword> function_name arg0 arg1
    echo $arg0
end
```
Or like this:
```
<function_keyword> function_name arg0 arg1 {
    echo $arg0
end
```
(See #99 for curly braces versus begin...end blocks)

The keyword could be `function`, `fn`, `def`, ...      
What do you think ?

skylerberg:
@nounoursheureux I like the first one. I am not picky on the keyword, but I like `fn` because it is less typing than function without any harm to readability in my opinion (everyone knows what `fn` is). I am excited to have functions that actually have named parameters in a shell language finally.

nounoursheureux:
@skylerberg so the function would be defined like this:
```
fn function_name arg0 arg1
    echo $arg0
end
```
And called like this:
```
function_name arg0="foo" arg1="bar"
```
@Ticki @jackpot51 are you OK with this ?

nounoursheureux:
I think we can close this

PR #105, opened 2016-03-03
Added functions without arguments

nounoursheureux:
Current syntax:
```
fn function_name
   echo "Hello, World"
end

function_name
> Hello, World
```

It works, but I feel that my code is quite bad (lots of `clone()` for example) so if you have any comments don't hesitate to tell me.       
Also, the `if...end` and `for...end` blocks don't work inside a function, so for now they are just aliases (but I'm going to try to fix this)

### DIFF
--- a/src/flow_control.rs
+++ b/src/flow_control.rs
@@ -6,9 +6,12 @@ pub fn is_flow_control_command(command: &str) -> bool {
     command == "end" || command == "if" || command == "else"
 }
 
+#[derive(Clone)]
 pub enum Statement {
     For(String, Vec<String>),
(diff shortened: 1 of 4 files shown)


nounoursheureux:
Related to #70

mmstick:
It's near-impossible to avoid all the instances of `clone()` with the usage of `if let` statements. `if let` statements are desugared into match statements whose conditions must all be alive at the same time, and because self is being used as the conditions, self is being borrowed multiple times. In order to not use `clone()`, you'd have to rewrite it as a series of basic self-contained `if` statements.

PR #106, opened 2016-03-03
[WIP] Support nested statements

nounoursheureux:
Currently there is no way to make an `if` inside another `if`, because the `end` instruction resets the current statement. To solve this problem I use a stack of `Statement`s: an `if` pushes a `Statement` to the stack and an `end` pops it. For now it doesn't work for the `for` blocks, because they only interpret their content when they encounter an `end` instruction, but I am working on it. Let me know if you agree with the approach I took so far !

Goals:
- [x] `if` inside `if`
- [ ] `if` inside `for`
- [ ] `for` inside `for`
...

### DIFF
--- a/src/flow_control.rs
+++ b/src/flow_control.rs
@@ -16,15 +16,21 @@ pub enum Statement {
 }
 
 #[derive(Clone)]
+pub enum Expression {
+    Job(Job),
+    Block(CodeBlock)
+}
+
+#[derive(Clone)]
 pub struct CodeBlock {
-    pub jobs: Vec<Job>,
+    pub expressions: Vec<Expression>,
     pub statement: Statement,
     pub collecting: bool,
 }
 
 impl CodeBlock {
     pub fn new(statement: Statement, collecting: bool) -> CodeBlock {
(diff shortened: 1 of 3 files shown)


mmstick:
It seems interesting to me and is at least something you can experiment with. How about the idea of nesting a function inside of a function?

nounoursheureux:
I removed the `Mode` struct and I replaced it with a variable in the `If` statement, I think it is cleaner. And I grouped the `CodeBlock` and `Job` structs in an `Expression` enum, so the blocks can hold other blocks and not only jobs. Let me know if you agree or not.
But I am a bit stuck: I think the appropriate data structure for the blocks hierarchy would be a tree: each block is a node and owns other blocks, and the jobs are the leafs. But I don't know how to implement this structure, because I need to have a pointer to the current block (to add new nodes/access its parent) and Rust won't let me read the tree if it is already borrowed . Maybe I'm thinking it wrong because this seems like a simple problem but I can't find any solution. Any ideas ?

nounoursheureux:
I am closing this as the codebase has changed a lot so there are lots of conflicts. I will try to submit smaller pull requests and have them merged one by one

PR #116, opened 2016-03-05
Ion Variables Patch

mmstick:
Several changes were necessary to allow Ion to use variables for tuning the shell. The `History` structure, for example, no longer needs to have a `max_size` variable because this information is already stored as a variable in the shell. This patch adds a few system variables such as `CWD`, `HISTORY_FILE`, `HISTORY_FILE_ENABLED`, `HISTORY_SIZE`, and `DIRECTORY_STACK_SIZE`. It also adds the capability for history being logged to a history file, named `ion_history`. What's currently lacking from this implementation is limiting the size of the history log and not logging commands loaded from files in history.

### DIFF
--- a/src/directory_stack.rs
+++ b/src/directory_stack.rs
@@ -1,11 +1,11 @@
 use std::collections::VecDeque;
 use std::env::{set_current_dir, current_dir, home_dir};
 use std::path::PathBuf;
+use variables::Variables;
 use super::status::{SUCCESS, FAILURE};
 
 pub struct DirectoryStack {
     dirs: VecDeque<PathBuf>, // The top is always the current directory
-    max_size: usize,
 }
 
 impl DirectoryStack {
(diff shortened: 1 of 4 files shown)


skylerberg:
Looks like the build is failing on Travis. Any clue what is up?

Ticki:
They probably changed the URL to the install script.

mmstick:
I'm not sure why it would be failing with Travis. It's having an error with a section of code that I haven't made changes to.

```
src/main.rs:117:13: 117:32 error: this pattern has 0 fields, but the corresponding variant has 1 field [E0023]
src/main.rs:117             Statement::Function => self.print_function_prompt(),
                            ^~~~~~~~~~~~~~~~~~~
```

skylerberg:
I am rerunning the last successful build of of master to see if it also fails: https://travis-ci.org/redox-os/ion/builds/113698354.

skylerberg:
Build is broken on master. I am looking into it.

skylerberg:
This is great! Very cool change. The one thing I am a little worried about is passing variables around all over the place, but I don't have a better idea off the top of my head. Maybe at some point in the future we can change that so that the function signatures make more sense.

jackpot51:
Shouldn't CWD be PWD?

stratact:
CWD == Current Working Directory
PWD == Present Working Directory

Same meaning different acronym. I guess, what makes more sense to stick with?

skylerberg:
If anyone has a strong reasoned opinion, I would love to hear it. I have no opinion on this.

jackpot51:
Bash, dash, ash, ksh use PWD. I'm not sure why, but they do, so other than asthetics I can't think of a reason to break convention

skylerberg:
Good enough for me, issue raised.

Ticki:
> I'm not sure why

POSIX, POSIX, POSIX.

PR #120, opened 2016-03-07
Add piping from one process to another

skylerberg:
This required some minor changes to the grammar which propagated through the rest of the code base. When using a pipeline builtin functions are ignored at this time. The logic for executing commands was updated into an undesirable state because of the complications with builtins. Piping logic was added  in a new file.


jackpot51:
Very cool!

PR #122, opened 2016-03-07
More flexible variable expansion

nounoursheureux:
You can now expand strings containing multiple variables (see the `expand_several_variables` test). The parser is incomplete for now (no escaping) and quite simplistic.


skylerberg:
Cool, this takes us one step closer to having a complete implementation of variables. I am going to merge this and then open a couple of issues for other similar improvements we could have for variable expansion.

PR #128, opened 2016-03-07
Escape string expansion with backslash / Fix #125

### LINKED ISSUE #125: Special variable `?` no longer expands
I didn't realize this until after merging https://github.com/redox-os/ion/pull/122, but since it only checks for alphanumeric and underscore it will not expand when you try `$?`.

nounoursheureux:
Fixes #125


skylerberg:
And you added tests too! Way to go!

Issue #131, opened 2016-03-07
Add io redirection operators


skylerberg:

jackpot51:
This is what I would like to see next, if possible :+1:

PR #137, opened 2016-03-08
Redirection

skylerberg:
I removed backgrounding for now because it allowed me to clean things up a lot and not have to deal with adding support for it to pipelines at the moment. I am opening several issues that this patch does not address.


skylerberg:
Fixes https://github.com/redox-os/ion/issues/131

PR #147, opened 2016-03-12
Add functions with arguments

nounoursheureux:
Syntax:
This creates a function called `print_two_strings` which takes two arguments:
```
fn print_two_strings first second
  echo $first $second
end
```
And this will call the function:
```
ion:# print_two_strings "Foo" "Bar"
Foo Bar
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -17,7 +17,7 @@ use self::peg::{parse, Job, Pipeline};
 use self::variables::Variables;
 use self::history::History;
 use self::flow_control::{FlowControl, is_flow_control_command, Statement};
-use self::status::SUCCESS;
+use self::status::{SUCCESS, NO_SUCH_COMMAND};
 use self::function::Function;
 use self::pipe::execute_pipeline;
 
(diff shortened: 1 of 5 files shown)


src/main.rs
```diff
@@ -208,14 +208,30 @@ impl Shell {
...
-        } else if self.functions.get(pipeline.jobs[0].command.as_str()).is_some() {
-            // Not really idiomatic but I don't know how to clone the value without borrowing self
-            let function = self.functions.get(pipeline.jobs[0].command.as_str()).unwrap().clone();
-            let mut return_value = None;
-            for function_pipeline in function.pipelines.iter() {
-                return_value = self.run_pipeline(function_pipeline, commands)
+        } else if let Some(function) = self.functions.get(pipeline.jobs[0].command.as_str()).cloned() {
+            if pipeline.jobs[0].args.len() - 1 != function.args.len() {
+                println!("This function takes {} arguments, but you provided {}", function.args.len(), pipeline.jobs[0].args.len()-1);
+                Some(NO_SUCH_COMMAND) // not sure if this is the right error code
```
skylerberg: Let's add another error code and pick an arbitrary int for it `INCORRECT_NUMBER_OF_ARGS` or something along those lines.

Issue #151, opened 2016-03-13
Glob expansion strips trailing slash characters


skylerberg:
If you run `echo ./` it will print `.` rather than `./`. This is because the pattern will be considered a glob and will be matched and then the match will not include a slash.

PR #152, opened 2016-03-13
Only expand globs if they contain a metacharater

skylerberg:


skylerberg:
Fixes: https://github.com/redox-os/ion/issues/151.

PR #162, opened 2016-03-21
Use drop instead of unlet

nounoursheureux:
Unlet isn't a real word, so it feels weird to use it to delete variables.

### DIFF
--- a/src/variables.rs
+++ b/src/variables.rs
@@ -61,7 +61,7 @@ impl Variables {
         SUCCESS
     }
 
-    pub fn drop_<I: IntoIterator>(&mut self, args: I) -> i32
+    pub fn drop_variable<I: IntoIterator>(&mut self, args: I) -> i32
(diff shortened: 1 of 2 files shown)


src/variables.rs
```diff
@@ -61,7 +61,7 @@ impl Variables {
         SUCCESS
     }
 
-    pub fn unlet<I: IntoIterator>(&mut self, args: I) -> i32
+    pub fn drop<I: IntoIterator>(&mut self, args: I) -> i32
```
Ticki: FYI, this might conflict with `Drop`, if ever implemented.

nounoursheureux:
@Ticki is it better with an underscore ?

Ticki:
Maybe `drop_variable`. Underscores are generally hacky, and might even make the confusion worse.

nounoursheureux:
@Ticki Agreed.

PR #170, opened 2016-04-17
Add the export command

nounoursheureux:

- Make `PWD`, `OLDPWD` and `HOME` environment variables
- Add the `export` command which sets environment variables.
Usage: `export KEY=VALUE`


polymetric1:
Wait isn't it supposed to be `let`?

Issue #168, opened 2016-03-28
Rewrite the tilde_expansion function without regular expressions


nounoursheureux:
**Problem**: Ion uses the `regex` crate for tilde expansion, which is not ported to Redox. Right now there is a `redox` branch without the `regex` crate and without tilde expansion, but this is just a workaround and not really a solution

**Solution**: We should rewrite the tilde expansion function without regular expressions

PR #171, opened 2016-06-04
rewrite tilde_expansion to remove regex dependency

MovingtoMars:
**Problem**: `tilde_expansion` uses the `regex` crate, which Redox doesn't support. See #168 

**Solution**: Rewrite `tilde_expansion` to remove `regex` dependency.

**Fixes**: #168

**State**: ready


jackpot51:
Thanks, this helps a lot!

PR #176, opened 2016-07-05
Rustify Error Handling

mmstick:
This change will merely modify error handling in a handful of situations to the methods provided by the `Option` and `Result` types for more functional-style error handling, which uses much less lines of code to achieve the same effect. It will also fix a few clippy warnings and makes changes to a few areas where files were being read directly into a `String::new()` without calculating the capacity beforehand, and values were being unnecessarily cloned.


skylerberg:
Cool!

PR #177, opened 2016-07-06
Liner

jackpot51:
Bring in liner, adding autocomplete and line editing to ion for Redox and Linux


MovingtoMars:
I think `liner` be specified by a specific commit in `Cargo.toml`.

This means we can bump it to force Cargo to get the new version of `liner` when there is an important upgrade (like MovingtoMars/liner#9).

Issue #123, opened 2016-03-07
Make variable expansion respect quoting


skylerberg:
Variables should be expanded during weak quoting (with double quotes) and they should not during hard quoting (with single quotes).

Issue #84, opened 2016-02-13
Don't expand variables in single quoted strings


skylerberg:
This should be changed after doing https://github.com/redox-os/ion/issues/83.

PR #178, opened 2016-07-09
Expose type of word to Variable logic via Word Enum

phddoom:
**Problem**: 

Single quoted words are currently expaned during variable expansion.

**Solution**: 

Add a mechanism to allow variable expansion logic to not expand single quoted words.

Introduces a Word enum with Bare, SingleQuoted, and DoubleQuoted
variants.

This allows logic related to variable expansion to not expand single
quoted words while expanding bare and double quoted words.

The Word Enum Derefs to String so it can access the functions / methods
of the string type.

**Drawbacks**: 

More clones

**TODOs**: 
- [x] Fix cargo test `PartialEq` error
- [x] Example test
- [x] Ensure this prevents all expansions not just variable
- [ ] Adjust variable expansion logic to search only for $[names of variables in scope]
- [ ] Remove commented code

**Fixes**: #123 #84

**State**: 

WIP

**Blocking/related**: #123 #84

**Other**:

I am seeking feedback on whether this is a good approach, or refinements. 

---

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -44,3 +44,10 @@ name = "unicode-width"
 version = "0.1.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 
+[metadata]
+"checksum glob 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)" = "8be18de09a56b60ed0edf84bc9df007e30040691af7acd1c41874faac5895bfb"
+"checksum libc 0.2.14 (registry+https://github.com/rust-lang/crates.io-index)" = "39dfaaa0f4da0f1a06876c5d94329d739ad0150868069cc235f1ddf80a0480e7"
+"checksum liner 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)" = "35e3e9a9a59faa8e143577b76173cb2b1225b77333b07ad27730269e70bcefd0"
+"checksum peg 0.3.14 (registry+https://github.com/rust-lang/crates.io-index)" = "2b5eb263593c6dde7cebf6422e655e1b7fddd0827721c00c05ac0c964ebed4da"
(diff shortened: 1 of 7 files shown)


skylerberg:
I have been thinking about this problem and the conclusion that I came to is that we need a separate lexing phase before parsing. The lexing phase should handle quoting, escaping, and maybe some or most of the types of expansions. It could also canonicalize whitespace to be a single space or make other such changes to make the parsing phase easier. This would require more clones, but it would decouple quoting/escaping from the grammar.

The problem with quoting/escaping in the grammar is that there are a lot of gnarly cases. As an example `echo 'a\sdf"'j${var}k\l"qwert"` contains a word that is partially double quoted, partially single quoted, a backslash that should be preserved while another should work as an escape character, and has a variable that should be expanded.

That said, I would accept changes that build on the grammar because they are still valuable improvements while we don't have a lexing phase. Going with that direction, this seems like a good way to do it.

Let me know your ideas and thoughts.

phddoom:
@skylerberg in your `echo 'a\sdf"'j${var}k\l"qwert"` example would the lexer produce tokens similar to:
```
echo
SingleQuotedWord a\sdf"
CONCAT
BareWord j${var}k\l
CONCAT
DoubleQuotedWord qwert
```
Where the lexer would then expand and concat where appropriate before sending to the parser?

Would we still want to use `rust-peg` for the lexer?

MovingtoMars:
@phddoom so `CONCAT` is used when two tokens are directly beside each other? Hmm, maybe the reverse woule be better. Have no extra token when two other tokens are touching, but have a `Whitespace` token when there is a space between tokens. IMO this method is simpler, and has the added advantage that it's easy to reconstruct the input string from the tokens.

mmstick:
We can close this now as it's long been resolved.

PR #180, opened 2016-07-21
update liner

MovingtoMars:
- bugfix in liner
- liner now specifies termion by commit to prevent breakage


jackpot51:
This is not possible to merge, we need @ticki 's  termion to modify its raw mode escape and then update the commit again.

jackpot51:
For now, use jackpot51/termion

PR #184, opened 2016-07-25
update to liner 0.1.1 and make full use of liner's History

MovingtoMars:
- use liner 0.1.1
- remove history.rs, use Context.History


jackpot51:
nice job!

PR #189, opened 2016-09-29
Improve variable brace expansion

mmstick:
This will fix the issue of variables not being expanded if there's more than one variable expansion within a word. It will also allow using colons to separate variable expansions, without requiring braces. Tests have also added to ensure that the functionality works.

```sh
$ let FOO = foo
$ let BAR = bar
$ echo $FOO:$BAR
foo:bar
$ echo ${FOO}:${BAR}
foo:bar
```

### DIFF
--- a/src/variables.rs
+++ b/src/variables.rs
@@ -231,85 +231,98 @@ impl Variables {
     fn brace_expand(&self, input: String) -> String {
         let mut ignore_next = false;
         let mut brace_found = false;
+        let mut variable_expansion_found;
         let mut infix_is_variable = false;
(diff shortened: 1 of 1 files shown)


jackpot51:
It is hard to see but it appears that this reads an empty env key at the end, in the case that it has a variable expansion like `echo ${FOO}:${BAR}`

This does not cause issues because the env functions typically disallow empty keys, but it may be possible in some cases to set an empty env key, causing ${FOO}:${BAR} to return, effectively ${FOO}:${BAR}${}

Issue #183, opened 2016-07-24
Tab completion of builtins and functions


ids1024:
Tab completion should support builtins and shell functions.

Edit: Variables should also be completed.

jackpot51:
working on it, thanks!

PR #191, opened 2016-10-23
Tab completion for functions and built-in commands

TomBebb:
This implements #183 by adding a `MultiCompleter` struct which combines several completers into one, and by adding a `BasicCompleter` with the function names and command names combined.


jackpot51:
Looks good to me - thanks!

PR #192, opened 2016-10-24
Use match rather than if/else if

ids1024:
Just a small style thing.

The constants defined in `status.rs` have also caught my attention. I'd have to look more closely at exactly how they are used to be sure, but it looks to me like they could be replaced by some kind of `enum`.

### DIFF
--- a/src/to_num.rs
+++ b/src/to_num.rs
@@ -16,16 +16,12 @@ impl ToNum for str {
 
         let mut num = 0;
         for c in self.chars() {
-            let digit;
-            if c >= '0' && c <= '9' {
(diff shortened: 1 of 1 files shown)


mmstick:
They could and should be an enum. Also, much of the functionality needs to be rewritten to pass back Results instead of printing errors immediately. Having errors handled properly would help a lot. There's a lot of work to be done. I'm currently working on improving brace expansion functionality to perfection, but it will take some time.

ids1024:
I'll take a shot at fixing that then. I can't guarantee anything though, since status is used fairly extensively.

PR #193, opened 2016-11-10
Brace/Variable Expansion v2

mmstick:
This is separating the brace/variable/etc shell expansion logic from the variables module into a new `shell_expand` module, which has modules of it's own. The solution takes advantage of custom `Iterators` to generate word and pattern-based tokens, and higher-order functions. A string is taken as an input, and is converted into an `Iterator` of pattern-matched words. These words are then mapped to their associated expansion actions.

This solution takes advantage of higher-order functions to retain modularity and to make it easier to write tests for. Braces are now also efficiently expanded such that they are properly permutated. It's passing all of the tests that I've written, but it could use much more extensive testing. A number of issues with the previous implementation is now fixed. This implementation should also be more efficient.

### Shell Expand Module
- mod.rs: provides the publicly-facing function for utilizing the module, and glues everything together.
- words.rs: takes the input string and converts it into an iterator of tokenized words
- braces.rs: contains functionality that expands brace words
- variables.rs: contains functionality that expands variable words

This module is then used by the existing `variables` module which just supplies the required functions to the string expansion function contained within to designate what shell-specific action needs to be applied for tilde and variable expansion.


mmstick:
Should be OK to merge now. I'm pretty much finished with making changes here.

jackpot51:
Thanks @mmstick!

PR #196, opened 2016-11-15
Improve Job Expansion & Better Error Handling

mmstick:
Provides a handful of changes to improve the `job_expand()` method in **variables.rs**, and some improvements to the **shell_expand** module in regards to error handling. Errors are now captured in the `words.rs` module instead of in the `braces.rs` module, which was causing a bug with improperly detecting some unmatched braces. Error messages will now also fully print the original command and correctly point out the location of the unmatched brace. An additional unit test was also added for the shell expand module.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -27,7 +27,6 @@ use self::status::{SUCCESS, NO_SUCH_COMMAND};
 use self::function::Function;
 use self::pipe::execute_pipeline;
 use self::shell_expand::ExpandErr;
-use self::shell_expand::braces::BraceErr;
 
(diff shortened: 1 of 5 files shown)


jackpot51:
This looks good - is it ready for merge?

mmstick:
This change is ready to merge. The next step will be to fix the issue with the peg parser eliminating spaces that are necessary.

jackpot51:
Do you think `process_expansion` is mergeable?

mmstick:
With a few changes, it should be mergeable.

PR #197, opened 2016-11-15
Process expansion

jackpot51:
Expand `$(command args)` by calling a subshell


mmstick:
Only change I'd make is to add a check for an unmatched parenthesis, similar to what's done for unmatched braces. It will be perfect after that.

PR #200, opened 2016-11-26
Add support for herestrings

klieth:
**Problem**: #133 

**Solution**: 

This adds a rule in the grammar for the <<< operator. It will capture a
word/quoted string after it and feed that literal string, including an
additional newline, to the stdin of the first process after it is
spawned.

**Drawbacks**: This combines the `execute_pipeline` and `pipe` methods, making a pretty big single method.

**Fixes**: #133

### DIFF
--- a/src/grammar.rustpeg
+++ b/src/grammar.rustpeg
@@ -72,11 +72,12 @@ redirection -> (Option<Redirection>, Option<Redirection>)
     / { (None, None) }
 
 redirect_stdin -> Redirection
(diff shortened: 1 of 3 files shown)


mmstick:
Recent changes to the grammar are causing a conflict this work.

klieth:
Yeah, I noticed there were some pretty massive changes. I'm going to close this for now; this approach isn't going to work given the rust-heavy approach taken with the parser.

PR #201, opened 2016-11-26
[Ready to Merge] Redirection & Process Expansion Fixes

mmstick:
Embarrassingly, a tiny bug in the last pull request caused redirection to not work. Process expansion also wasn't working when that process contained a semicolon. This change will fix that. It's now possible to use commands like the following:

```sh
echo $(ls > temp; cat temp | grep Cargo)
```

### DIFF
--- a/src/grammar.rustpeg
+++ b/src/grammar.rustpeg
@@ -121,9 +121,11 @@ _pipelines -> Statement
                     },
                     '>' => {
                         mode = RedirMode::Stdout;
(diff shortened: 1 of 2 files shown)


mmstick:
With this change, I will now begin using ion as my default shell on Linux.

jackpot51:
Awesome work!

jackpot51:
@mmstick If you send an email to info@redox-os.org, I can add you to the developer chat

PR #202, opened 2016-11-29
Convert Vec to Pipeline Iterator

mmstick:
This change will convert the existing pipeline collector in the grammar into an `Iterator` to eliminate heap allocations caused by pushing to a `Vec`.

### DIFF
--- a/src/grammar.rustpeg
+++ b/src/grammar.rustpeg
@@ -1,6 +1,7 @@
 use peg::Pipeline;
 use peg::Job;
 use peg::Redirection;
+use peg::PipelineIterator;
 use flow_control::Statement;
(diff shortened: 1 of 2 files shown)


mmstick:
Seems Travis is having some issues.

```
failed to fetch `https://github.com/rust-lang/crates.io-index`
```

mmstick:
[Found the cause](https://www.reddit.com/r/rust/comments/5fjf7o/psa_cargo_nightly_on_linux_cant_fetch_the_index/). Cargo nightly can't fetch the index and the fix will be released tomorrow.

jackpot51:
It looks good, so I will merge. The fix

PR #203, opened 2016-11-29
Quoted Variables Retain Newlines

mmstick:
This will fix the issue of newlines not being retained when a variable is quoted.

Example:

```
let A = "$(seq 1 10)"; echo $A; echo "$A"
1 2 3 4 5 6 7 8 9 10
1
2
3
4
5
6
7
8
9
10
```


jackpot51:
Looks good - lots of good changes recently!

PR #204, opened 2016-11-30
Bit Twiddling in Shell Expansion

mmstick:
The shell expansion logic contains a large number of boolean values, each which will consume 8 bits of memory and an 8-bit register even though booleans are logically 1-bit values. By collapsing all the booleans into a single 8-bit integer and using bit-twiddling operations, shell expansion should be much faster and use less memory.

This also allows me to refactor some code in shell expansion to keep Clippy quiet.


jackpot51:
Cool

PR #205, opened 2016-11-30
Bit-Twiddle PipelineIterator & Clippy Fixes

mmstick:
This will perform the boolean bit-twiddling optimization to the PiplelineIterator which reduces six 8-bit fields into one u8 and allows for more optimal conditional checking by checking multiple values at the same time. There's also some clippy fixes in here.

With this submission, I'm done for the day.

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -78,14 +78,16 @@ impl Shell {
                     let (words, pos) = editor.get_words_and_cursor_position();
 
                     let filename = match pos {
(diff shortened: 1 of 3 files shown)


jackpot51:
Thanks for all the PRs!

mmstick:
A small bug slipped through from this pull request which wasn't caught by any of the existing tests so I might submit the fix shortly alongside a test to prevent the bug it in the future.

jackpot51:
Sounds good, thanks

PR #206, opened 2016-11-30
Fix Last Pull Request

mmstick:
Fixed an issue from the last pull request and added test units for the following two cases:

```
let A = "$(seq 1 10)"; echo $A; echo "$A"
```

```
let A = $(seq 1 10); echo $A; echo "$A"
```


jackpot51:
Thanks, glad to see those test cases

PR #211, opened 2017-02-02
Remove unncessary attribute #![feature(deque_extras)]

polymetric1:
this fixes the build on the latest rust nightly build 2017-01-31. It's apparently no longer needed. Not really sure why it counts as a full-on build stopping error, though.


jackpot51:
Thanks

PR #213, opened 2017-02-12
Expand variables in if statements (fixes #212)

### LINKED ISSUE #212: if statement does not work with variables
We need to handle variable names in conditions so we can compare them. The if statement implementation is pretty simple right now. Do you have any other idea what could be implemented and how to improve it?

**Reproduction**: 
Add debug message:
```diff
diff --git a/src/main.rs b/src/main.rs
index d614815..aa2d50a 100644
--- a/src/main.rs
+++ b/src/main.rs
@@ -260,6 +260,7 @@ impl Shell {
     }
 
     fn handle_if(&mut self, left: String, comparitor: Comparitor, right: String) {
+        println!("DBG: {:?} == {:?}", left, right);
         let value = match comparitor {
             Comparitor::GreaterThan        => { left >  right },
             Comparitor::GreaterThanOrEqual => { left >= right },
```
Run this script:
```txt
let test=0

if $test == 0
    echo pass
else
    echo fail
end
```

**Expected behavior**: 
pass

**Actual behavior**: [describe the actual behavior, which is presented through the repro.].
```txt
cargo run examples/variable_number_comparison.ion
    Finished dev [unoptimized + debuginfo] target(s) in 0.0 secs
     Running `target/debug/ion examples/variable_number_comparison.ion`
DBG: "$test" == "0"
fail
```

fcapano:
Added expansion of variables when evaluating if statements.
Undefined variables will evaluate to empty strings, thus the equality of two undefined variables will evaluate to True.
This is the same behaviour of other shells such as Bash, Sh, Zsh, ...

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -260,6 +260,17 @@ impl Shell {
     }
 
     fn handle_if(&mut self, left: String, comparitor: Comparitor, right: String) {
+
+        let left = match self.variables.expand_string(&left, &self.directory_stack) {
+            Ok(ref expanded_string) => { expanded_string.clone() },
+            Err(_) => { "".to_string() },
+        };
+
+        let right = match self.variables.expand_string(&right, &self.directory_stack) {
+            Ok(ref expanded_string) => { expanded_string.clone() },
+            Err(_) => { "".to_string() },
+        };
+
         let value = match comparitor {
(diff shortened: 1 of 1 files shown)


jackpot51:
Thanks!

Issue #214, opened 2017-02-17
Space pipe (aka the Arkanoïd bug)


Mojo4242:
Space matters with pipe, it probably should not:

**Reproduction**
```
ion:file:/root# echo 12345 | cat
12345
ion:file:/root# echo 12345 |cat
12345
ion:file:/root# echo 12345| cat

ion:file:/root#
```

**Expected behavior**
```
ion:file:/root# echo 12345 | cat
12345
ion:file:/root# echo 12345 |cat
12345
ion:file:/root# echo 12345| cat
12345
ion:file:/root#
```
----
**rustc -V** :
```
rustc 1.17.0-nightly (29dece1c8 2017-02-08)
```

**rustup show** :
```
Default host: x86_64-unknown-linux-gnu

nightly-x86_64-unknown-linux-gnu (default)
rustc 1.17.0-nightly (29dece1c8 2017-02-08)
```

**git rev-parse HEAD** :
```
9ea162c920a9a65719754482fd59121fe74175dc
```

**qemu-system-x86_64 -version** :
```
QEMU emulator version 2.5.0 (Debian 1:2.5+dfsg-5ubuntu10.8), Copyright (c) 2003-2008 Fabrice Bellard
```

**uname -a** :
```
Linux 74ae79d00b85 4.9.8-moby #1 SMP Wed Feb 8 09:56:43 UTC 2017 x86_64 x86_64 x86_64 GNU/Linux
```

mmstick:
Easily fixed. It will just require a single line to be changed. [This one](https://github.com/redox-os/ion/blob/962c3ec9e004a48bad9ed6913c7097a4286ba3e5/src/peg.rs#L281).

```rust
    ...
    '|' if (flags & IS_VALID == 0) => job_found!(false),
    ...
```

The `IS_VALID` constant is set to:

```rust
const IS_VALID: u8 = 255 ^ (BACKSLASH + WHITESPACE + COMMENT);
```

So basically pipelines should be changed to use this instead:

```rust
255 ^ (BACKSLASH + COMMENT)
```

mmstick:
This can be closed

jackpot51:
Sweet

PR #216, opened 2017-03-02
Pipeline Fix & Avoid Pointless UTF8 Checks

mmstick:
This will fix the spacing issue with pipelines, and it will also
address the issue of non-ASCII characters being processed wrong,
and improving performance by disabling UTF8 checks where they are
redundant.

Instead of iterating over characters to check for specific ASCII
characters, this change will have strings iterated by bytes, which
allows strings to be correctly spliced when there's non-ASCII
characters in the mix. This is faster as bytes don't require UTF8
validation, and most text will consist of ASCII characters. It will
also correctly slice strings at the correct index so that strings
will not be butchered.

There is a use of the unsafe keyword when splicing strings in some
sections of the code to convert a vector of bytes into a String
without checking, but this is safe because the source where the
string originates was already UTF8 valid in the first place. These
sections are merely adding or removing specific ASCII characters.

Fixes issue #214


jackpot51:
Looks good - thanks!

PR #217, opened 2017-03-02
For Grammar Improvements

mmstick:
This change will connect the for grammar to shell_expand, thus enabling
the possibility to write expressions like so:

```fish
for index in $(seq 0 10)
    echo $index
end
```

This also implements inclusive and exclusive ranges. Not sure if we want to go with this syntax, but I figured I'd use the same syntax as Rust.

### Inclusive Range

```fish
for index in 0...10
    echo $index
end
```

### Exclusive Range

```fish
for index in 0..11
    echo $index
end
```

Eliminates the need to use the `seq` command.

### DIFF
--- a/src/flow_control.rs
+++ b/src/flow_control.rs
@@ -1,4 +1,8 @@
 use super::peg::Pipeline;
+use directory_stack::DirectoryStack;
+use variables::Variables;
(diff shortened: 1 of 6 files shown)


jackpot51:
Great work!

PR #218, opened 2017-03-03
Split Statements

mmstick:
This makes it possible to now write commands like so:

```fish
for index in $(seq 1 10); echo $index; end
```


jackpot51:
Thanks!

PR #219, opened 2017-03-04
Parser Refactoring, Unit Tests Added, & New Features

mmstick:
### Refactoring

This will primarily refactor all the parsing-related content into a new parser module. The pipeline parsing logic was removed from the peg module and placed into it's own module.

### Unit Tests

Unit tests were added for the new `for` grammar and statement splitter, so there's more coverage now. A number of tests were removed for the pipleline logic as it's no valid since landing the statement splitter.

### StatementSplitter Improvements

The statement splitter was improved to now ignore comments, as it should. The following command:

```sh
echo $(echo one #two); echo three # four
```

Will thus split into:

```sh
echo $(echo one #two)
echo three
```
(Seems highlight.js doesn't properly parse comments)

The statement splitter is now also integrated with the rest of the shell's functionality (scripts, init script, source, etc.) instead of just the REPL shell alone.

### Process Recursion

It's now possible to perform process recursion, like so:

```sh
echo $(echo $(echo one))
```

### Context History Improvement

In addition, a small change was appended that improves the usefulness of navigating history when a previous command contained multiple statements, which also disables the context history for commands executed via scripts (as other shells do it).


mmstick:
Heads up: about to land support for process recursion

jackpot51:
This is really cool - ion is becoming a shell I can use by default

mmstick:
Later today I'll have a fix for single quotes within process expansions. Seems you can't write commands like so:

```sh
echo $(git branch | rg '[*]')
```

mmstick:
And then it seems work needs to be done to allow for nested for loops.

jackpot51:
Sounds good, thanks @mmstick!

PR #220, opened 2017-03-05
Parser Fixes

mmstick:
With this change, the parsing bugs have been resolved, and it's now possible to write commands like the following:

```sh
let current_branch = $(git status -b | awk 'NR == 1 {print $3}')
let modified = $(git status | rg 'modified' -c)
echo $(echo one $(echo two) three)
```


jackpot51:
Thanks very much!

Issue #215, opened 2017-02-20
Command aliases


jeandudey:
Will command aliases be supported?

jeandudey:
About that, trying to workaround this with `fn` gives this bug:

#### Example
```
fn ls
ls --color
end
```
in shell:
```
ion:/home/jeandudey/# ls
This function takes 0 arguments, but you provided 1
```

I don't know if this is a bug, or a intentional feature (recursiveness?).

msehnout:
I think the intention is to use functions for this purpose.

```txt
ion:/# ll
ion: command not found: ll
ion:/# fn ll
fn> ls -l
fn> end
ion:/# ll
total 20
...
```

mmstick:
The `alias` command simply hasn't been implemented yet. You're free to implement it if you want though.

mmstick:
I'm working on this at the moment.

mmstick:
I'm halfway through implementing this. Syntax will expand upon the existing `let` functionality by adding a new `alias` keyword to define that the following binding will be an alias binding versus a variable binding.

```sh
let alias ls = 'ls --color'
```

mmstick:
This can be closed now. You can write aliases like so:

```sh
alias ls = 'ls --color'
ls -l
```

jackpot51:
@mmstick I also added `unalias`

PR #222, opened 2017-03-05
Implement Aliasing Support

mmstick:
Implements support for creating aliases, resolving issue #215. I've implemented the syntax to utilize the existing `let` keyword, like so:

```sh
let alias ls = "ls --color"
ls -l
# Expands to and executes `ls --color -l`
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -65,7 +65,6 @@ impl Default for Shell {
 impl Shell {
     fn readln(&mut self) -> Option<String> {
         let prompt = self.prompt();
-
         let funcs = &self.functions;
         let vars = &self.variables;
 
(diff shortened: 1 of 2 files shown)


jackpot51:
I don't like the `let alias` syntax. Would it be difficult to change it to `alias`?

Also, I wonder what happens if you have this:
```
let alias echo = "echo alias"
echo echo
```
Will it correctly echo `alias echo`, or incorrectly echo `alias echo alias`

mmstick:
> I don't like the let alias syntax. Would it be difficult to change it to alias?

I could make it `alias` if you would prefer that.

> Will it correctly echo alias echo, or incorrectly echo alias echo alias?

Indeed, the implementation will expand `echo echo` into `echo alias echo`

jackpot51:
Good to know! I would prefer `alias`, to avoid multiplexing `let` syntax

mmstick:
Okay, it's been changed to use an `alias` keyword instead.

jackpot51:
Thanks @mmstick!

PR #223, opened 2017-03-06
Write Errors to Stderr & History Completion

mmstick:
- Errors are now written to standard error instead of standard out
- Command history is now available as completions for liner
- Fixed variables in single quotes being expanded in subshells


jackpot51:
Is there a reason to return Cow::Borrowed instead of just a &'static str ?

mmstick:
The return value may be a `String` instead of a `&'static str`.

jackpot51:
Looks good then.

PR #227, opened 2017-03-07
Implement `&&` and `||` Operators

mmstick:
Resolves issue #227 by implementing support for `&&` (**AND**) and `||` (**OR**) operators. There's also basic background support, so background jobs will run in the background and print when they are complete.

The following command:

```sh
echo one && echo two && echo three
```

Will print

```
one
two
three
```

While the following:

```sh
echo one && echo two || echo one
```

Will print

```
one
two
```

And the following:

```sh
ecsdf one && echo two && echo three || echo four
```

Will print

```
ion: command not found: ecsdf
four
```

### DIFF
--- a/src/flow_control.rs
+++ b/src/flow_control.rs
@@ -17,12 +17,12 @@ pub enum Statement {
     },
     Else,
     End,
+    // TODO: Vec is unnecessary here because there will always be one pipeline parsed
     Pipelines(Vec<Pipeline>),
     Default
 }
 
 impl Statement {
-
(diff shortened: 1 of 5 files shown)


mmstick:
I need to quickly implement a fix before this is merged.

jackpot51:
Thanks @mmstick! Any ideas on how to fuzz the grammer parser?

mmstick:
Not that I know of, besides continuing to add more and more unit tests, and adding a more comprehensive suite of scripts to test against.

PR #228, opened 2017-03-08
Fix : src/main.rs : 103 file_completers for all directories in ${PATH}

CapitanCrazy:
Sorry for my awful english

**Problem**: I add some code for add completions from all ${PATH} directories;

- I rewrite ```MultiCompleter``` to use ```Vec<FileCompleter>``` instead of ```FileCompleter```

### DIFF
--- a/src/completer.rs
+++ b/src/completer.rs
@@ -1,13 +1,13 @@
 use liner::Completer;
 
 /// A completer that combines suggestions from multiple completers.
-#[derive(Clone, Copy, Eq, PartialEq)]
+#[derive(Clone, Eq, PartialEq)]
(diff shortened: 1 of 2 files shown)


jackpot51:
PATH on Redox and Windows is seperated by `;`, not `:`

jackpot51:
Should be if cfg!(unix)

CapitanCrazy:
Ok

jackpot51:
Thanks! Now it will work on all systems that Rust works on!

mmstick:
Probably should have removed the TODO notice since you implemented the feature.

PR #229, opened 2017-03-08
Implement While Support

mmstick:
The current implementation is simple in that it will parse the expression for each loop and check if the value is either **true** or **1**. I plan to give it the ability to perform conditional expressions of it's own later when I have more time.

It doesn't have to be merged now if you'd prefer to wait for the conditional expressions support.


jackpot51:
That's ok - I like to have the latest and greatest.

PR #234, opened 2017-03-09
Refactor Let/Alias/Export & Implement Builtin README

mmstick:
- The let/alias/export/drop commands have been refactored into a builtins module
- A README.md file was created for documenting the builtins module and the commands contained within.

### DIFF
--- a/src/builtins/README.md
+++ b/src/builtins/README.md
@@ -1,0 +1,48 @@
+# Ion Shell Builtins
+
+This directory contains the source code of Ion's builtin commands and documentation for their usage.
+
+## Variables
+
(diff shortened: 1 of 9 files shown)


mmstick:
Fixing tests -- will be finished soon.

mmstick:
And finished

jackpot51:
I think this expands twice - let does not need to expand on its own

mmstick:
I've removed the additional expansions, and added a README.md for the builtins module.

mmstick:
I'm going to do a but more refactoring to cut down the required SLOC further than it's already been pruned.

jackpot51:
Thanks, looks good. Can you suggest tests to catch double expand behavior?

mmstick:
I could probably make it a compile-time error.

PR #235, opened 2017-03-10
Implement Proper While Support

mmstick:
With these changes, the while loops in Ion will work as while loops in other shells -- executes a supplied command and checks if the exit status was a success value, and quits when it's not.

This means that you can do this in Ion now:

```ion
let a = 1;
while test $a -lt 10
    echo $a
    let a = $(echo "$a + 1" | bc)
end
```

And this:

```ion
while read line
    echo $line
end
```

Although with the second example, you'll be stuck in an endless loop as Ctrl+C / Ctrl+D doesn't work.


jackpot51:
Interesting. Could `if` be changed to parse the following item as a command as well?

mmstick:
It will require a bit more effort to pull off than the work for `while`, but I plan to look into it later.

jackpot51:
Thanks very much @mmstick !

PR #236, opened 2017-03-10
Implement Basic Arithmetic For Let/Export

mmstick:
WIth this change, it is possible to take variables which contain numbers, and apply basic arithmetic to them using the let command, without the need for an external application. It performs 32-bit floating point math, so irrational numbers are supported.

```ion
let a = 1
echo $a
let a += 4
echo $a
let a *= 10
echo $a
let a /= 2
echo $a
let a -= 5
echo $a
```

```ion
let a = 1;
while test $a -lt 100
    echo $a
    let a += 1
end
```


mmstick:
Going to re-open this without the previous pull request attempting to merge.

PR #239, opened 2017-03-12
[RTM] Flow Control Rewrite

mmstick:
### Checklist

- [x] While Loops
- [x] For Loops
- [x] Nested Loops
- [x] If Conditions
- [x] Nested If Conditions
- [x] Else If Conditions
- [x] Functions
- [x] Re-implement Aliasing Support
- [x] General Code Cleanups / Fix TODOs
- [x] Handle Leftover Statements
- [x] Update Tests / Examples


mmstick:
Travis is broken at the moment, but I've now achieved the milestone of passing all tests. Added a few additional example tests to test for inclusive, for exclusive, nested for loops, and while loops.

mmstick:
All the functionality I planned to implement is now finished. All that's left is handling leftover statements and general code cleanup here and there.

mmstick:
The work is now complete. The flow control logic has been redone, and you can now use it to the fullest extent.

- Flow control statements contain all of the statements that they will potentially execute
- Partially formed statements are buffered in memory and executed as soon as they are complete
- Nesting works for all flow control statements, so you can write a function that contains a function, or loops that contain loops
- If conditions now work properly the same way while conditions work
- Else if conditions are now supported -- any number of else if conditions may be provided
- The code detects if an else if is supplied after an else or if an else supplied after an else and stops collection and prints an error as it should.

jackpot51:
@mmstick Very good work, and it looks like there are tests for much of the behavior. At this point I want to increment ion to 1.0. I have given you write access to ion, which you should have had before

mmstick:
As a fun fact, you can now add something like this to your `ionrc` and perform a recursive fibonacci function.

```ion
fn fib x
    if test $x -lt 2
        /usr/bin/echo $x
    else
        let a = $(fib $(/usr/bin/echo "$x - 1" | bc))
        let b = $(fib $(/usr/bin/echo "$x - 2" | bc))
        /usr/bin/echo "$a + $b" | bc
    end
end
```

Have to use `/usr/bin/echo` though because builtin commands don't support pipes.

jackpot51:
That is cool. Should `ion` have a way to list fn defines?

mmstick:
Can't hurt to list function definitions too.

PR #267, opened 2017-04-18
Echo builtin

paezao:
**Problem**: Adding coreutils/echo as a ion builtin command

**Solution**: I have added the full coreutils/echo source code as a builtin command and changed it slightly to work with ion API

- Added coreutils as a dependency to make use of ArgParser
- Added a new echo.rs to src/builtins containing the echo functionality
- Added a command at src/builtins/mod.rs to call the new echo function
- Added echo command to src/shell/job.rs making it possible to pipeline

**State**: ready

### DIFF
--- a/src/builtins/echo.rs
+++ b/src/builtins/echo.rs
@@ -1,5 +1,10 @@
 use std::io::{self, Write, BufWriter};
 
+const HELP: u8 = 1;
+const ESCAPE: u8 = 2;
+const NO_NEWLINE: u8 = 3;
+const NO_SPACES: u8 = 4;
+
 const MAN_PAGE: &'static str = /* @MANSTART{echo} */ r#"
 NAME
     echo - display a line of text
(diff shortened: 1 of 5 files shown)


src/builtins/echo.rs
```diff
@@ -0,0 +1,122 @@
...
+
+pub fn echo(args: &[String]) -> Result<(), io::Error> {
+    let mut parser = ArgParser::new(4)
+        .add_flag(&["e", "escape"])
+        .add_flag(&["n", "no-newline"])
+        .add_flag(&["s", "no-spaces"])
+        .add_flag(&["h", "help"]);
+    parser.parse(args.iter().cloned());
+
+    let mut buffer: BufWriter<Box<Write>> = BufWriter::new(Box::new(io::stdout()));
```
mmstick: To avoid the need to Box, you can write it like this:

```rust
let stdout = io::stdout();
let mut buffer = BufWriter::new(stdout.lock());
```

src/builtins/echo.rs
```diff
@@ -0,0 +1,135 @@
...
+        }
+    }
+
+    let stdout = io::stdout();
+    let mut buffer = BufWriter::new(stdout.lock());
+
+    if arguments.contains(&"help") {
+        buffer.write_all(MAN_PAGE.as_bytes())?;
+        buffer.flush()?;
+        return Ok(());
```
mmstick: Why not move this up to and replace line 37?

src/builtins/echo.rs
```diff
@@ -0,0 +1,135 @@
...
+        if arg == "--help" {
+            arguments.push("help");
+        } else if arg.starts_with("-") {
+            let arg = &arg[1..];
+            for argopt in arg.chars() {
+                match argopt {
+                    'e' => arguments.push("escape"),
+                    'n' => arguments.push("no-newline"),
+                    's' => arguments.push("no-spaces"),
+                    'h' => arguments.push("help"),
```
mmstick: Could just have a single u8 represent all possible options, like so:

```rust
const ESCAPE: u8 = 1;
const NO_SPACE: u8 = 2;
const NO_NEWLINE: u8 = 4;

let mut flags = 0u8;
flags |= ESCAPE; // set the escape flag
if flags & (255 ^ ESCAPE) != 0 {
    // Executes if the escape flag was set
}
```

This also doesn't seem like it would function with long form options which are denoted with two -'s instead of one.

src/builtins/echo.rs
paezao: Because it needs to work with the short form as well.

src/builtins/echo.rs
```diff
@@ -0,0 +1,146 @@
+use std::io::{self, Write, BufWriter};
+
+const HELP: u8 = 1;
+const ESCAPE: u8 = 2;
+const NO_NEWLINE: u8 = 3;
```
mmstick: Values have to be base 2, so 1, 2, 4, 8, etc.

src/builtins/echo.rs
```diff
@@ -0,0 +1,146 @@
...
+                    data.push(arg);
+                }
+            }
+        }
+    }
+
+    let stdout = io::stdout();
+    let mut buffer = BufWriter::new(stdout.lock());
+
+    if (flags & HELP) != 0 {
```
mmstick: Bit flags are checked by comparing the AND of the reversed bits of the flag, so `if flags & (255 ^ HELP) != 0`.

PR #271, opened 2017-04-23
Increase amount of work done without allocation

jFransham:

We allocate a lot

Use `SmallVec` and `SmallString` (based on `SmallVec`) to allow small vectors and strings to be stored on the stack.

Some public API functions that took `String`s take `SmallString`s now. Also, it's now even faster on my machine.

- [...]
- [...]
- [...]

**Drawbacks**: [if any, describe the drawbacks of this pull request.]

Clean up the code, unify the methods of creating a singleton smallvec (probably with a macro), move the typedefs to a `types` module.

**Fixes**: [what issues this fixes.]

WIP

**Blocking/related**: [issues or PRs blocking or being related to this issue.]

**Other**: [optional: for other relevant information that should be known or cannot be described in the other fields.]

------

_The above template is not necessary for smaller PRs._

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,8 @@ liner = "0.1.4"
 peg-syntax-ext = "0.4"
 permutate = "0.2"
 unicode-segmentation = "1.1"
+smallvec = "0.3.3"
+smallstring = { path = "src/smallstring" }
 
 [target.'cfg(all(unix, not(target_os = "redox")))'.dependencies]
 users = "0.5.1"
--- a/src/ascii_helpers.rs
+++ b/src/ascii_helpers.rs
@@ -1,0 +1,45 @@
+use std::ascii::AsciiExt;
+use std::ops::DerefMut;
+
+// TODO: These could be generalised to work on non-ASCII characters (and even
+//       strings!) as long as the byte size of the needle and haystack match.
+
+pub trait AsciiReplaceInPlace {
+    fn ascii_replace_in_place(&mut self, needle: char, haystack: char);
+}
+
+pub trait AsciiReplace: Sized {
+    fn ascii_replace(self, needle: char, haystack: char) -> Self;
+}
+
+impl<T: DerefMut<Target=str>> AsciiReplace for T {
+    fn ascii_replace(mut self, needle: char, haystack: char) -> Self {
+        self.ascii_replace_in_place(needle, haystack);
+        self
+    }
+}
+
+impl AsciiReplaceInPlace for str {
+    // I tried replacing these `assert!` calls with `debug_assert!` but it looks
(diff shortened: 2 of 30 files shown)


mmstick:
Just noticed the pull request. It's great to see more work done on allocating less! Anyway, I pushed some changes to let assignments earlier which have a few conflicts with these changes at the moment.

jFransham:
@mmstick The usage of small strings isn't a huge win, but I've just pushed a change to replacements that knocks about 15% off the fibonacci benchmark time on my computer

jFransham:
Profiling the fib benchmark shows that a lot of time is spent round-tripping to and from int, but I think that on average workloads this is unlikely to be as important.

mmstick:
Would you like to attempt to merge this work as is?

jFransham:
It's pretty messy, but it is a win. I tried doing the in-place replace change without the small string work (because the latter doesn't seem to make a huge difference) but it seems to slow it down a fair bit. Adding back the small strings (I made a new branch and cherry-picked the replacement work, so I just switched back to master) brings the speed boost back. I have no idea why this is.

If I get it cleaned up it can be merged as-is for a nice speed boost. I've made most of the code able to handle either `SmallString` or `String` and there's a type synonym that determines which to use, so it's pretty easy to roll back the changes if necessary.

jFransham:
Added a `FromIterator` impl for `SmallString` that knocks another 5% off the runtime (assuming my computer is as fast now as it was when I tested it before, I've got the same programs running and I was on power both times but we'll have to verify the speedup). This includes a custom-built `Utf8Iterator` struct, I couldn't find anything like it in the stdlib but if there is a prebuilt solution I should use that instead.

If you've got no objections to the code as it is now, I'll squash the commits and we can get this sucker merged.

jFransham:
Woah! I just rebased on master and got a 1s speed boost! I don't know what did it but Ion is now overall more than 30% faster than dash!

jFransham:
So based on the fibonacci benchmark from your /r/rust post, 2000 iterations, average of 100 runs each, I get:

```
Dash:
  min: 7.09
  max: 8.37
  avg: 7.22
Ion:
  min: 4.46
  max: 4.98
  avg: 4.66
```

By my math, that's a consistent 35% speedup on this particular benchmark.

mmstick:
Thanks! I have merged the work.

jFransham:
Do you mind me posting to /r/rust? I'm excited about such a significant speedup compared to Dash, since it was only recently that I was considering writing an optimising POSIX sh replacement but talked myself out of it because I considered Dash too mature to top.

mmstick:
You may if you want to.

mmstick:
And I'm certain that there is still room for improvement.

PR #292, opened 2017-06-06
Correctly parse statements with multiple escaped characters

### LINKED ISSUE #291: Filenames with a complete set of parentheses are not expanded properly
**Reproduction**: Create a filename such as `foo()` and attempt to refer to it using escaped parentheses.

**Expected behavior**: The following completes with no output, implying no error has occurred.
```
$ touch "foo()"
$ cat foo\(\)
```
**Actual behavior**:
```
$ touch "foo()"
$ cat foo\(\)
cat: 'foo(\)': No such file or directory
```

**Build information**: 
- `rustc -V => rustc 1.19.0-nightly (14f30da61 2017-05-21)`
- `get rev-parse HEAD => 3352a0cb8c67da3aff135ad7849b366d35bcd407`

**Blocking/related**: 
- I came across this issue while confirming that #284 works as expected
- #4 covers this, but it is still open

**Misc**: 
- I'm guessing (based on some minimal prodding) the issue has to do with the behavior of `WordIterator` inside `parser::shell_expand::expand_string`. The contents of [`token_buffer` at this point](https://github.com/redox-os/ion/blob/3352a0cb8c67da3aff135ad7849b366d35bcd407/src/parser/shell_expand/mod.rs#L135) ends up being: `[Normal("foo"), Normal("(\\)")]` where it should be `[Normal("foo"), Normal("("), Normal(")")]`
- Confirmed above, submitting PR for review shortly
```

huntergoldstein:
**Problem**: Statements such as `foo\(\)` were not being correctly parsed, see #291.

**Solution**: Whenever a character is read by a `WordIterator` when the `BACKSL` flag is set, the flag gets unset.

- Backslash flag now reset upon reading the escaped character
- Add regression test for multiple backslashes

**Drawbacks**:
- It could use some more tests for these sorts of use cases
- It is unclear if this affects the other uses of the `WordIterator` beyond parsing command line statements.

**Fixes**: Resolves #291.
**State**: Ready with the caveat of adding more unit / regression tests, also formatting and such.

### DIFF
--- a/src/parser/shell_expand/words.rs
+++ b/src/parser/shell_expand/words.rs
@@ -932,4 +932,19 @@ mod tests {
         ];
         compare(input, expected);
     }
+
+    #[test]
+    fn test_multiple_escapes() {
+        let input = "foo\\(\\) bar\\(\\)";
+        let expected = vec![
+            WordToken::Normal("foo"),
+            WordToken::Normal("("),
+            WordToken::Normal(")"),
+            WordToken::Whitespace(" "),
+            WordToken::Normal("bar"),
+            WordToken::Normal("("),
+            WordToken::Normal(")"),
+        ];
+        compare(input, expected);
+    }
 }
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks for the pull request.

Issue #258, opened 2017-03-31
Carefully Glob What Needs Globbing


mmstick:
Globbing should happen during shell expansion, instead of after. To do this, only normal words that have valid globbing signatures should be expanded in the shell expansion module. Some work may have to be done to check whether a bracket belongs to an array or a glob.

PR #295, opened 2017-06-12
Moving the glob parser

mgmoens:
**Problem**:Globbing during shell expansion (see https://github.com/redox-os/ion/issues/258)

**Solution**: I added an extra param to the Normal token which designates it as a normal token that's globable or not.

- Globbing gets moved and probably (isn't) yet parsed 100% correctly. I'm just submitting this PR so I can get in touch with contributors and notify other people of my attempt at trying to fix this.

**Drawbacks**:

Wildcards during quotes don't parse correctly.

Make sure wildcards in quotes get parsed as text (if this happened during the previous version).

**Fixes**: [what issues this fixes.]
https://github.com/redox-os/ion/issues/258
Ready for review
**Blocking/related**: [issues or PRs blocking or being related to this issue.]

**Other**: The author isn't all that used to programming in Rust.

### DIFF
--- a/src/parser/shell_expand/words.rs
+++ b/src/parser/shell_expand/words.rs
@@ -705,6 +705,7 @@ impl<'a> Iterator for WordIterator<'a> {
                         let mut rewind = true;
                         let mut iter = iterator.clone();
                         while let Some(character) = iter.next() {
+                            moves+=1;
                             match character
                             {
                                 b'[' => {
@@ -714,23 +715,21 @@ impl<'a> Iterator for WordIterator<'a> {
                                     break;
                                 },
                                 b']' => {
-                                    if moves<=2 && square_bracket == 1 {
+                                    if moves<=3 && square_bracket == 1 {
                                     }
                                     else{
                                         rewind = false;
-                                        moves+=1;
                                         break;
                                     }
                                 }
                                 _=> (),
                             }
-                            moves+=1;
                         }
                         if rewind == false{
                             for _ in 0..moves {
                                 iterator.next();
                             }
-                            self.read+=moves;
+                            self.read+=moves+1;
                             glob = true;
                             break;
                         }
@@ -830,6 +829,7 @@ impl<'a> Iterator for WordIterator<'a> {
                     let mut rewind = true;
                     let mut iter = iterator.clone();
                     while let Some(character) = iter.next() {
+                        moves+=1;
                         match character
                         {
                             b'[' => {
@@ -839,24 +839,21 @@ impl<'a> Iterator for WordIterator<'a> {
                                 break;
                             },
                             b']' => {
-                                if moves<=2 && square_bracket == 1 {
-                                    break;
+                                if moves<=3 && square_bracket == 1 {
                                 }
                                 else{
-                                    moves+=1;
                                     rewind = false;
                                     break;
                                 }
                             }
                             _=> (),
                         }
-                        moves+=1;
                     }
                     if rewind == false{
                         for _ in 0..moves {
                             iterator.next();
                         }
-                        self.read+=moves;
+                        self.read+=moves+1;
                         glob = true;
                     }
                     else {
--- a/src/parser/shell_expand/words.rs
+++ b/src/parser/shell_expand/words.rs
@@ -201,7 +201,7 @@ impl<'a> ArrayMethod<'a> {
             }
         }
 
-       
+
         Some("".into()).into_iter().collect()
     }
 }
@@ -219,6 +219,7 @@ pub enum WordToken<'a> {
     Process(&'a str, bool, Index),
     StringMethod(&'a str, &'a str, &'a str, Index),
     ArrayMethod(ArrayMethod<'a>),
+    Glob(&'a str),
 }
 
 pub struct WordIterator<'a> {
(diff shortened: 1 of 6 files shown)


mmstick:
Thanks for working on this. It's one of the biggest issues with Ion at the moment.

mgmoens:
Is there an IRC channel type thing where you guys hang around? I'd like to ask a couple of questions regarding the code.

mmstick:
I'd be the one to contact regarding Ion, as I wrote all the code here and therefore know all the ins and outs of the implementation. That said, the rest of the Redox team can be found on the team Mattermost chat service: https://chat.redox-os.org/

Or you can ask questions here on this PR or the associated issue, or [send me a direct email](mailto:mmstickman@gmail.com).

Something to look out for with glob expansion is that it will need to return multiple strings, rather than a single string. If glob expressions are found, each expansion should be a unique element so that they work correctly with for loops. The for loop code will also have to be slightly modified, as it's currently using a workaround hack that expands globs specifically for for loops.

mgmoens:
Yup, I thought it was nested globbing that didn't work but it's actually printing out multiple filenames at once, nested globbing works if there's only one dir with one files.

Also, what does join(" ") do? I see that it can be invoked on Small Vecs but it's not in the crates docs.

mmstick:
The `join` method applies to Vec/Array types that contain strings. They just concatenate them with a pattern. In the case of `join(" ")`, that would be joining all strings with a space between each element.

mgmoens:
Found a way to fix the return multiple strings part. 

For the bracket globbing (ie [A-Z]argo.toml => Cargo.toml), can I consider the text inside brackets a glob token if there are no spaces between brackets, no double dots and the string contained is not a single character?

mmstick:
I believe the approach we can use is that if a single-element array does not contain spaces after the [ and before the ], then it is a glob.

mgmoens:
There's still a bug where starting with [] makes the last character spawn it's own string for no reason. The character get's read twice, cus the glob is correct.

I tried using the pipes with examples that did work correctly and it seemed to work fine. Imma add some comments.

Why is Travis failing?

mgmoens:
Ok, I think I'm done. I think it's reviewable at the moment. Fixed the bug I was talking about.

src/parser/shell_expand/words.rs
```diff
@@ -201,14 +201,14 @@ impl<'a> ArrayMethod<'a> {
...
-       
+
         Some("".into()).into_iter().collect()
     }
 }
 
 #[derive(Debug, PartialEq, Clone)]
 pub enum WordToken<'a> {
-    Normal(&'a str),
+    Normal(&'a str,bool),
```
mmstick: Wouldn't it be better to return `Glob(&'a str)` instead?

src/parser/shell_expand/words.rs
```diff
@@ -755,27 +806,71 @@ impl<'a> Iterator for WordIterator<'a> {
...
+                    if rewind == false{
+                        for _ in 0..moves {
+                            iterator.next();
+                        }
+                        self.read+=moves+1;
+                        glob = true;
+                    }
+                    else {
+                        return Some(WordToken::Normal(&self.data[start..self.read],glob));
+                    }
```
mmstick: Should be able to make the above a new function that returns a `Some(&'a str)` so it doesn't need to be copied. Then you can use the map method to convert the returned `Some(&'a str)` into the correct token that you need.

src/shell/job.rs
```diff
@@ -37,10 +37,10 @@ impl Job {
             let mut iterator = self.args.drain();
             expanded.push(iterator.next().unwrap());
             for arg in iterator.flat_map(|argument| expand_string(&argument, expanders, false)) {
-                if arg.contains(|chr| chr == '?' || chr == '*' || chr == '[') {
+                /*if arg.contains(|chr| chr == '?' || chr == '*' || chr == '[') {
```
mmstick: Maybe remove all the commented code now?

mmstick: Perhaps also add a test in the **examples** directory that checks if the following functions correctly? All you do is create the ion file that contains the commands to run, and the corresponding out file that will have the correct output of the standard output of the ion script when execute.

```
echo br*.*
> braces.ion braces.out break.ion break.out
echo [ef]*.ion
> else_if.ion else_if.out fail.ion fail.out fibonacci.ion fibonacci.out fn.ion fn.out for.ion for.out
```

src/parser/shell_expand/words.rs
mgmoens: I did it this way to not have an extra if else construct in basically every place that returns a Normal. If you insist I'll gladly change it but I think it's gonna mess up the code.

src/shell/job.rs
mgmoens: Added. It's in examples/glob.ion and examples/glob.out.

mgmoens: Done.

src/parser/shell_expand/words.rs
mgmoens: I don't really get what you're saying. Basically, in the match (character) part I peek ahead to see if the contents of the string inside the outer [] is globable. If it isn't, I just redirect it to the code that was there before. I know I copy the iterator and that it isn't the most efficient thing, but I need a way to look ahead without break the iterator you use in the next(..) function.

mmstick: What I'm referring to is that you have this logic copied in two different places. One here that ends at line 868, and another copy that ends at line 743.

mmstick:
I'll merge this as is for now.

mgmoens:
I was not available the last couple hours, I'll split up the code regarding 868 etc. tomorrow.

mmstick:
The splitting won't be necessary as I've already done that. Thanks anyway for the PR request! I know this wasn't a particularly easy feature to work on, in addition to getting adjusted to the codebase.

PR #297, opened 2017-06-13
Add semantics for indexing into arrays with negative indices 

### LINKED ISSUE #263: Improve Range Syntax With Negative Values for Taking From the End
To further improve the capabilities of the range syntax, we can expand on that to support the following syntax, where a negative value indicates to count and take values from the end of an array, versus from the beginning

```ion
# Syntax for getting the last value?
echo @array[-1]   # Take the next to last value
echo @array[..-1] # Take all but the last value
echo @array[-5..] # Take the last five values
echo @array[-5..-2]
```

huntergoldstein:
**Problem**: There is no way to refer to, for example, the second to last element of an array aside from `@array[len(array) - 2]` where `len` gets the number of elements in an array.

**Solution**: Extend Ion's syntax to allow for this through negative indices, such as in Python:
```
$ let array = [ 1 2 3 4 5 6 ]
$ echo @array[-1]
6
$ echo @array[1..-2]
2 3 4
```

- Introduce new data structures that represent indexing from the end of an array
- Update parsing of array ranges and indices to use these new structures

**Drawbacks**:
- `parser::shell_expand::words` has become a bit of a mess with a bunch of `Index` named classes everywhere. As discussed in #263 it would be useful to refactor this and move some of the logic repeated at call-sites, but that's a little too broad for a single PR.

**TODOs**: 
- [x] Outstanding bug where `@array[2...-2]` results in an empty output for arrays where it is well defined.
- [x] Add in more regression / unit tests, including failure-like cases

**State**: WIP

**Blocking/related**: Closes #263

### DIFF
--- a/src/builtins/variables.rs
+++ b/src/builtins/variables.rs
@@ -238,7 +238,7 @@ pub fn export_variable<'a, S: AsRef<str> + 'a>(vars: &mut Variables, args: &[S])
 #[cfg(test)]
 mod test {
     use super::*;
-    use parser::{expand_string, ExpanderFunctions, Index, IndexEnd};
+    use parser::{expand_string, ExpanderFunctions, Index};
     use shell::status::{FAILURE, SUCCESS};
     use shell::directory_stack::DirectoryStack;
 
--- a/src/parser/loops/for_grammar.rs
+++ b/src/parser/loops/for_grammar.rs
@@ -1,7 +1,7 @@
 use shell::directory_stack::DirectoryStack;
 use shell::variables::Variables;
 use types::Value;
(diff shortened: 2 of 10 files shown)


huntergoldstein:
This is ready for review; There is some cleanup to be done but I'm not sure how I feel about saying "It will come with refactoring the `Index` stuff."

mmstick:
It looks good to me. If you want it merged now, I can do that.

huntergoldstein:
Merge away

mmstick:
Thanks! I'm sure it was a lot of work getting adjusted to the codebase.

PR #298, opened 2017-06-15
Parse ² and ³ as one would expect in the calc command

mgmoens:
**Problem**: ² and ³ are obvious mathematical characters (my terminal even changes ^2 into ²), but triggered errors in calc.

**Solution**: Added two specific tokens, square and cube, to the calc tokens. 

- Added these tokens and added some code in the exponentiation fn.

**Drawbacks**: I think none.

**TODOs**: Maybe write a test?

**Fixes**: None. I was playing around in calc and noticed this.

**State**: Ready.

**Blocking/related**: I think none.

**Other**: None

------

### DIFF
--- a/src/builtins/calc.rs
+++ b/src/builtins/calc.rs
@@ -9,6 +9,8 @@ pub enum Token {
     Divide,
     Multiply,
     Exponent,
+    Square,
+    Cube,
     OpenParen,
     CloseParen,
     // TODO: Don't pass around a string when we can pass around a number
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks

mmstick:
Think you'd be willing to also implement:

- Bitwise AND (&)
- Bitwise OR (|)
- Bitwise NOT (~)
- Bitwise XOR (^)
- Bitwise Right Shift (>>)
- Bitwise Left Shift (<<)
- Modulo (%)

Only problem is that XOR is normally defined with the `^` character, but we are using it as the exponentiation operator. We should change the exponentiation operator to `**` instead. This is what Bash uses for exponentiation.

```bash
$ echo $((0 ^ 1))
1
```

These operators should also only be allowed when the number is a whole number.

mgmoens:
Yup, I'm gonna (try to) implement these.

PR #302, opened 2017-06-15
Refactor `Index`, etc. into `Select`, `Range`, and `Index`

huntergoldstein:
**Problem**: The extra features introduced in #297 made it clear that `Index` `IndexStart` and `IndexEnd` were confusing to exist all together. 

**Solution**: This PR consolidates and appropriately separates the responisbilities of `Index`, `IndexStart`, and `IndexEnd` into `Select`, `Range`, and `Index.

- Replaced top-level `Index` with `Select` which better captures the meaning of the enumeration: a selection of an array versus an index, which brings to mind a single element within the array.
- Added a separate `Range` struct that encapsulates inclusive vs. exclusive
- Consolidated `Index<blah>` into `Index::Forward` and `Index::Backward` and removed all of the `CatchAll` values. `CatchAll` is now represented by `Forward(0)` or `Backward(0)`.
- Added explicit, versus implicit, failure cases when a range is invalid

**Drawbacks**: 
- `Selection` is a more apt name but I wonder if there's something more precise
- This is mostly due to the features in #297 but there are several places where some logic is repeated

**TODOs**:
- [ ] Move some of the repeated logic and patterns for iterators from their call-sites to `Range`; I would love it if this PR reduced the total line count 
- [x] Add documentation and real examples for `Range` and `Index`
- [x] Add in more regression tests, especially now that `Range` is easily independently testable

**State**: Good to go

**Related**: #297, #263

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -3,34 +3,34 @@ macro_rules! get_expanders {
     ($vars:expr, $dir_stack:expr) => {
         ExpanderFunctions {
             tilde: &|tilde: &str| $vars.tilde_expansion(tilde, $dir_stack),
-            array: &|array: &str, index: Index| {
+            array: &|array: &str, selection : Select| {
                 use std::iter::FromIterator;
                 use $crate::types::*;
 
                 match $vars.get_array(array) {
-                    Some(array) => match index {
-                        Index::None   => None,
-                        Index::All    => Some(array.clone()),
-                        Index::ID(id) => array.get(id).map(
-                            |x| Array::from_iter(Some(x.to_owned()))
-                        ),
-                        Index::FromEnd(idx) => {
-                            if array.len() < idx {
-                                None
-                            } else {
-                                array.get(array.len() - idx)
-                                     .map(|x| Array::from_iter(Some(x.to_owned())))
-                            }
+                    Some(array) => match selection {
+                        Select::None  => None,
+                        Select::All   => Some(array.clone()),
+                        Select::Index(id) => {
+                            id.resolve(array.len())
+                              .and_then(|n| array.get(n))
+                              .map(|x| Array::from_iter(Some(x.to_owned())))
                         },
-                        Index::Range(start, end) => {
-                            let len = array.len();
-                            let array : Array =
-                                array.iter()
-                                     .skip(start.resolve(len))
-                                     .take(end.diff(&start, len))
+                        Select::Range(range) => {
+                            if let Some((start, length)) = range.bounds(array.len()) {
+                                let array = array.iter()
+                                     .skip(start)
+                                     .take(length)
                                      .map(|x| x.to_owned())
-                                     .collect::<_>();
-                            if array.is_empty() { None } else { Some(array) }
+                                     .collect::<Array>();
+                                if array.is_empty() {
+                                    None
+                                } else {
+                                    Some(array)
+                                }
+                            } else {
+                                None
+                            }
                         }
                     },
                     None => None
@@ -59,7 +59,7 @@ pub mod shell_expand;
 mod statements;
 mod quotes;
 
-pub use self::shell_expand::{Index, IndexEnd, ExpanderFunctions, expand_string, expand_tokens, WordToken, WordIterator};
+pub use self::shell_expand::{Select, Range, Index, ExpanderFunctions, expand_string, expand_tokens, WordToken, WordIterator};
 pub use self::arguments::ArgumentSplitter;
 pub use self::loops::for_grammar::ForExpression;
 pub use self::statements::{StatementSplitter, StatementError, check_statement};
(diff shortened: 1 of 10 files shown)


src/parser/mod.rs
```diff
@@ -3,34 +3,34 @@ macro_rules! get_expanders {
...
-                                array.get(array.len() - idx)
-                                     .map(|x| Array::from_iter(Some(x.to_owned())))
-                            }
+                    Some(array) => match selection {
+                        Select::None  => None,
+                        Select::All   => Some(array.clone()),
+                        Select::Index(id) => {
+                            id.resolve(array.len())
+                              .and_then(|n| array.get(n))
+                              .map(|x| Array::from_iter(Some(x.to_owned())))
```
huntergoldstein: There are a bunch of places where `ion` uses strange syntax in order to create a singleton `SmallVec`. Maybe its time to add in a `smallvec![...]` macro?

src/parser/mod.rs
```diff
@@ -3,34 +3,34 @@ macro_rules! get_expanders {
...
+                                     .take(length)
                                      .map(|x| x.to_owned())
-                                     .collect::<_>();
-                            if array.is_empty() { None } else { Some(array) }
+                                     .collect::<Array>();
+                                if array.is_empty() {
+                                    None
+                                } else {
+                                    Some(array)
+                                }
```
huntergoldstein: @mmstick Is there a reason that this function will map the empty `Array` to `None`? I don't see this anywhere else.

mmstick: I'd have to look into this code more deeply as I didn't write this part, but I believe it's to make the distinction that an empty array has no value, and therefore is `None`, likely for better pattern matching later down the line.

src/parser/mod.rs
mmstick: If you have any ideas for refactoring that would make the code more readable/manageable, you're welcome to pursue them. These were put in place by @jFransham to mitigate memory allocations in the shell and thus achieve even better performance compared to Dash.

src/parser/mod.rs
huntergoldstein: That is what it looks like: also looks like there ends up being no cosmetic difference as the empty array is rendered the same as no output.

src/parser/shell_expand/ranges.rs
```diff
@@ -123,10 +120,14 @@ pub fn parse_index_range(input: &str) -> Option<(IndexStart, IndexEnd)> {
...
-    assert_eq!(Some((IndexStart::new(2), IndexEnd::FromEnd(1))), parse_index_range("2...-2"));
-    assert_eq!(Some((IndexStart::new(0), IndexEnd::FromEnd(0))), parse_index_range("0...-1"));
+    let range1 = Range::exclusive(Index::Forward(0), Index::Forward(3));
+    let range2 = Range::inclusive(Index::Forward(0), Index::Forward(2));
+    let range3 = Range::inclusive(Index::Forward(2), Index::Backward(1));
+    let range4 = Range::inclusive(Index::Forward(0), Index::Backward(0));
+    assert_eq!(Some(range1), parse_index_range("0..3"));
+    assert_eq!(Some(range2), parse_index_range("0...2"));
+    assert_eq!(Some(range3), parse_index_range("2...-2"));
+    assert_eq!(Some(range4), parse_index_range("0...-1"));
```
mmstick: What of `parse_index_range("-3..-1")` and `parse_index_range("-3..")`?

huntergoldstein: Good catch; see `c5fa695`.

huntergoldstein:
Notes about the first "Todo", refactoring some of the common `Range` / `Select` logic:
- My strategy has been to attempt to write some member method of `Select` or `Range` like:
```rust
fn apply<T, I>(&self, I : iter) -> Box<Iterator<Item=T>> where I : Iterator<Item=T> {
  ...
}
```
This requires some funky type lifetimes where `T` has to be a "static" type, which might not be the case. Also note that the return value is _required_ to be boxed as `Iterator` has an unknown size at compile time. All of the selections except `Select::All` return a new iterator versus modifying the old one. This could be solved with some macro magic, maybe.
- There are a couple call-site specific tricks that are used for dealing with an `Index::Backward` or getting the length (Can we clone the Iterator? Do we just manually re-make it? Is it an `ExactSizeIterator`) which would, most likely, involve multiple methods.
- Overall I'm not sure how much moving the logic to `Select` would add; it might just be more confusing.

mmstick:
The length of the iterator should be known at runtime by taking the size of the array and doing a calculation like `array.len() - start - (end - start)`

huntergoldstein:
That works for arrays but there are a few different places where `Select` is used. I could probably just replace those callsites yeah.

mmstick:
Merged your pull request. Thanks for the work!

Issue #255, opened 2017-03-29
Move Let / Export Into Grammar


mmstick:
This is required in order to move forward with arrays as first class citizens in the shell, because it is vital to not expand the let statement before parsing it.

```ion
let array = [ one two three ]
```

It is also required in order to pass complex arithmetic expressions and have the assignment parser distinguish between an arithmetic expression and regular expression, in the event that we would like to gain support for such a feature.

```ion
let a = (b * h) ^ 2
```

- [x] Migrate let into grammar
- [ ] Migrate export into grammar
- [x] If parsed let statement returns a vector, create an array
- [ ] Implement basic mathematics for arrays

```ion
$ let a = 1
$ let a += [ 2 2 5 ]
$ echo $a
10
$ let b = [ 2 2 5 ]
$ let b += 5
$ echo @b
7 7 10
```

PR #306, opened 2017-06-17
Move `export` into the grammar

huntergoldstein:
**Problem**: `export` not being in the grammar meant that the RHS was eagerly expanded.

**Solution**: `export` now has the same semantics as `let` but instead of storing the variable locally, it is written to the OS environment.

- `export` is now in the PEG grammar and has an associated `Statement` enumeration
- `export` and `let` now have the same logic except, as mentioned, `export` writes to the OS environment
- DRYed up the logic, including replacing a custom made `ExpanderFunctions with `get_expanders!`

**Drawbacks**: Nothing off the top of my head 

**Fixes**: Related to #255

**State**: Ready!

### DIFF
--- a/src/builtins/mod.rs
+++ b/src/builtins/mod.rs
@@ -8,7 +8,7 @@ mod echo;
 mod calc;
 mod set;
 
-use self::variables::{alias, drop_alias, drop_variable, export_variable};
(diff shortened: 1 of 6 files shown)


mmstick:
Thanks

Issue #262, opened 2017-04-06
Implement Braced Arrays


mmstick:
As there is a braced variant of string variables, it would be nice to also have a braced variant for arrays:

```ion
@{array}
@{array[0..3]}
```

PR #309, opened 2017-06-19
Fix braced variables not parsing correctly

huntergoldstein:
At some point in the history of `StatementSplitter` it started rejecting braced variables such as `${foo}`. While working on #262 I noticed this.

The problem was that the first case, which catches any character that cannot be in a braced variable, also threw an error for a brace, even though that might be the valid end of a braced variable.


mmstick:
Thanks for noticing this!

PR #312, opened 2017-06-20
Replace manual bit flags with auto-generated bit flag construct

### LINKED ISSUE #308: Use bitflags instead of home-rolled constants for flags
# Problem
- Rules and bit fiddling can be difficult to remember for bitflags; the more someone has to think through and remember how this all works the more likely they are to make a mistake
-  It is harder to debug which flags are set when: most methods either are prone to human error or are tedious

# Solution
We could make use of the [bitflags crate](https://docs.rs/bitflags/0.9.1/bitflags/) which should be as efficient as just using constants but increases quality of life (see above).

huntergoldstein:
**Problem**: There are small ergonomics issues with using hand-rolled bit flags as described in #308.

**Solution**: Replace the use of the hand-rolled flags with auto-generated `bitflags` structures that provide nice mnemonics for membership while being mostly compatible with the old syntax.

- Bitflags have been replaced with `bitflag!` equivalents as per the [bitflag crate](https://docs.rs/bitflags/0.9.1/bitflags/)

**Drawbacks**: 
- I'm not entirely sure if this is any more readable than before: part of the problem is `contains` versus `intersects`, where `foo.contains(bar)` is `foo & bar == bar` and `foo.intersects(bar)` is `foo & bar != 0`. These names are non-intuitive.
- In addition to the above, the operators tended to be much more concise so for the most part I tried to not replace them with the mnemonics. 
- There is a _slight_ chance that this is overall less efficient (more function calls) but I highly doubt it; the `bitflags` crate essentially has `#[inline]` on every function

**Fixes**: Closes #308.

**State**: I'm on the fence on whether or not this is a worthwhile addition, but I have nothing else to add.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,7 @@ name = "ion-shell"
 version = "1.0.3"
 dependencies = [
  "ansi_term 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "bitflags 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "fnv 1.0.5 (registry+https://github.com/rust-lang/crates.io-index)",
  "glob 0.2.11 (registry+https://github.com/rust-lang/crates.io-index)",
  "liner 0.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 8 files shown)


mmstick:
Thanks

PR #317, opened 2017-06-20
Hide Tokio crates behind cfg gate for Redox

### LINKED ISSUE #314: Fix Redox Build
tokio-signal brings in a large number of dependencies, many that do not support Redox yet. Using a lighter weight signal handling crate, or not linking to tokio-signal when compiling for Redox would help.

huntergoldstein:
Closes #314.

I'm still trying to set up a redox environment (see: https://github.com/redox-os/redox/issues/970; I'm giving up trying to do this locally for now, spinning up Ubuntu VM) but in the meantime this should work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,7 +20,6 @@ name = "ion"
 [dependencies]
 bitflags = "0.9.1"
 fnv = "1.0"
-futures = "0.1"
 glob = "0.2"
 liner = "0.2"
 permutate = "0.3"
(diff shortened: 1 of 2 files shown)


jackpot51:
This looks good to me. Can you also change `Cargo.toml` ?

Cargo.toml
```diff
@@ -27,6 +27,8 @@ permutate = "0.3"
 unicode-segmentation = "1.2"
 smallvec = "0.4"
 smallstring = "0.1"
+
+[target.'cfg(not(target_os = "redox"))'.dependencies]
```
jackpot51: Should this also contain `futures` ?

huntergoldstein: It should and it does now.

mmstick:
Thanks! This should work for handling Redox.

PR #327, opened 2017-06-22
[WIP] Add arithmetic sub-expressions to Ion's syntax

### LINKED ISSUE #243: Arithmetic Expressions
In the future, it should be possible to perform arithmetic by containing such expressions within `$((...))`, such as how Bash performs it. As an example:

```bash
let result = $(( ($x * $y) * 2 ))
```

huntergoldstein:
**Fixes**: Closes #243 

- Builtin `calc` functionality is now exposed to the rest of the crate
- Added support for the `$((...))` syntax form which is equivalent to `calc '...'` with some syntactic sugar for referring to variables

**Drawbacks**: 
- The `calc` builtin is now tightly coupled to the rest of the crate: I don't feel this is desirable

**TODOs**: 
- [x] Add tests for parsing at `statement` and `word` level
- [x] Revise flags and logic in `parser::statements`; this could be cleaner IMO

**State**: Core functionality should be there, needs some polish

### DIFF
--- a/src/builtins/calc.rs
+++ b/src/builtins/calc.rs
@@ -473,7 +473,7 @@ pub fn parse(tokens: &[Token]) -> Result<String, CalcError> {
     d_expr(tokens).map(|answer| answer.value.to_string())
 }
 
-fn eval(input: &str) -> Result<String, CalcError> {
+pub fn eval(input: &str) -> Result<String, CalcError> {
     tokenize(input).and_then(|x| parse(&x))
 }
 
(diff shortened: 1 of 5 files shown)


mmstick:
Very nice implementation that you have for arithmetic. Whenever you are ready, we can merge this work. I would like to eventually split the logic up in `parser::statements` into diverging nano-parsers.

huntergoldstein:
Ready to merge!

PR #338, opened 2017-06-23
Add `@graphemes(..)`, `@bytes(..)` and `@chars(..)`

### LINKED ISSUE #316: Grapheme & Byte-based String Iterations
We also need to come up with a syntax for iterating a string by graphemes, and possibly byte-based iterations too. I think we could use string methods that return arrays for this:

```ion
for grapheme in @graphemes(string)
    echo $grapheme
end

for char in @chars(string)
    echo $char
end

for byte in @bytes(string)
    echo $byte
end
```

huntergoldstein:
**Problem**: Users cannot iterate over the contents of a string easily.

**Solution**: Add in three new `ArrayMethod`s that map from string variables to arrays:
- `@graphemes(..)` iterates over the unicode graphemes of the string as per [this crate](https://github.com/unicode-rs/unicode-segmentation)
- `@bytes(..)` iterates over the UTF-8 byte sequence that corresponds to the given string
- `@chars(..)` iterates over the unicode scalars of the byte sequence as per [`String::chars`](https://doc.rust-lang.org/std/string/struct.String.html#method.chars)

- Add the three above methods to `parser::shell_expand::words`
- Introduce a new trait for applying a `Select` variant to an iterator DRYer

**Drawbacks**: 
- The `SelectWithSize` trait is not optimal as it always requires the size of the iterator that is passed in; in most cases we are not dealing with `ExactSizeIterator` so this requires making a copy of the iterator and using `Iterator::count()`.
- Having `SelectWithSize::select` work like `collect` means that it is costly to map / filter / reduce after calling `select`, but I don't think this will be an issue.

**Fixes**: Closes #316 

**State**: WIP: I'm not quite sure if `SelectWithSize` is the right approach. Any and all thoughts are appreciated.

**Other**: 
- Depending on how we feel about `SelectWithSize` I want to start refactoring code that can be replaced with a similar pattern.
- The alternative for `SelectWithSize` is a `Select<I>` iterator; we could implement `From<I>` for other iterators, which would allow us to specialize the logic more efficiently (versus having a catch-all solution that).

### DIFF
--- a/src/parser/shell_expand/words.rs
+++ b/src/parser/shell_expand/words.rs
@@ -1324,4 +1324,49 @@ mod tests {
         compare(input, expected);
     }
 
+    macro_rules! functions_with_vars {
+        () => {
+            ExpanderFunctions {
+                vars:     &Variables::default(),
+                tilde:    &|_| None,
+                array:    &|_, _| None,
+                variable:  &|var : &str, _| match var {
+                    "pkmn1" => "Pokémon".to_owned().into(),
+                    "pkmn2" => "Poke\u{0301}mon".to_owned().into(),
+                    _ => None
+                },
+                command: &|_, _| None,
+            }
+        }
+    }
+
+    #[test]
+    fn array_methods() {
+        let expanders = functions_with_vars!();
+        let method = ArrayMethod {
+            method: "graphemes",
+            variable: "pkmn1",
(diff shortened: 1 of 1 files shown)


mmstick:
I think your `SelectWithSize` solution is fine for now. We're free to experiment with different solutions in the future. Thanks for the work!

PR #341, opened 2017-06-24
Refactor `array_expand` (and squash bug)

huntergoldstein:
- Refactored `array_expand` to be the entry point for any `Select`ion into an array of words, DRYing up the code along the way
- Squashed a subtle bug where reverse indexing into an array that expanded to a larger array (for example the array `@[[foo bar] [baz quux]]` would incorrectly use the original length of the array and not the expanded length. This has been resolved by getting a reversed iterator over the array instead of using the size to resolve the true length.
- Added a regression test for the above

**State**: Good to go!

**Other**: The majority of my time spent on this PR was trying to figure out the best way to expand `Select` code reuse:
- I'm probably overthinking this. Just having functions like `array_expand` work beautifully; they aren't as pretty as `foobar.select(..)` but they work and are as efficient as possible, generally.
- It will be much easier if I pass around `Box<Iterator>`s as I don't need to commit to a particular iterator type and instead use composition of iterators. I have noticed there is / was a clear effort to reduce the use of heap allocated structures as well as references to removing the allocation increasing performance. Are there benchmark tests lying around that I could make use of?

### DIFF
--- a/src/parser/shell_expand/mod.rs
+++ b/src/parser/shell_expand/mod.rs
@@ -81,14 +81,10 @@ fn array_expand(elements : &[&str], expand_func: &ExpanderFunctions, selection :
 }
 
 fn array_nth(elements: &[&str], expand_func: &ExpanderFunctions, index: Index) -> Option<Value> {
+    let mut expanded = elements.iter().flat_map(|e| expand_string(e, expand_func, false));
     match index {
-        Index::Forward(n) => elements.iter()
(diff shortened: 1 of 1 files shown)


mmstick:
The main benchmark I've used is the Fibonacci script located in the examples directory, upped to 200 loops, and benchmarked / profiled with the perf command.

Anyway, thanks for the work!

PR #342, opened 2017-06-25
Shell::readln now completes to a filename if the last word is a file 

### LINKED ISSUE #321: restrict ./ expansion to filenames
`./f<Tab>` shows all commands beginning with 'f', and it shows files starting with 'f'; it should only show the files

huntergoldstein:
**Related Issues**: Closes #321 

**Changes introduced in this pull request**:
- `Shell::readln` now determines if the first word is either an existing directory, or is potentially a directory (i.e. if the parent exists as a folder)

**TODO**:
- [x] Use `DirectoryStack` instead of getting the current directory from the environment
- [x] How does this react if the user types a completely wrong directory?

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -77,7 +77,7 @@ impl<'a> Shell<'a> {
             sigint_handle: ctrl_c
         }
     }
-
+    
     fn readln(&mut self) -> Option<String> {
         let vars_ptr = &self.variables as *const Variables;
         let dirs_ptr = &self.directory_stack as *const DirectoryStack;
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -77,6 +77,7 @@ impl<'a> Shell<'a> {
             sigint_handle: ctrl_c
         }
     }
+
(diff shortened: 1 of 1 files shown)


huntergoldstein:
It seems like its easier to just use `env::current_dir` instead. @mmstick is there a reason we are passing around raw pointers in `Shell::readln`? Typing in a completely wrong directory seems to just offer no completions, so this is good to go for me.

mmstick:
Due to the way completions work, there is no way to perform our Ion-specific completions using safe Rust. That would perhaps require some big changes to the way liner works. Completions happen within the completions method, but are only executed by liner when stored within the `Shell`'s `Context`'s `completer` field, and that means that in order to use our `tilde_expansion` function, the shell will therefore need to effectively contain references to itself, which can only be done with raw pointers.

Our `tilde_expansion` method requires access to both access to `Variables` for the `PWD` and `OLDPWD` variables in the shell, and access to `DirectoryStack` for obtaining values off the directory stack when a number is passed with the tilde character, hence why we need access to the fields in the first place.

Anyway, thanks for the work!

PR #346, opened 2017-06-27
Empty string can now be passed in as an argument to a function / command / etc.

### LINKED ISSUE #345: empty string arguments
the empty string is not treated as a valid argument:

```
# touch a
# rename '' 0 a
rename: not enough arguments
# fn foo x; end
# foo ''
ion: function takes 1 arguments, but you provided 0
```

huntergoldstein:
- Allow the empty string to be parsed as `WordToken::Normal("")` by allowing a double quote to be the first char of a `WordToken::Normal` sequence
- Added a case where if a `WordToken::Normal` would be the empty string, replace it with whatever the next token would be
  - This prevents cases like `"$FOO"` from generating empty tokens due to having a quote immediately before a `$`
- Added a regression test for this behavior

**Fixes**: Closes #345.

**State**: Good to go!

### DIFF
--- a/src/parser/shell_expand/words.rs
+++ b/src/parser/shell_expand/words.rs
@@ -923,10 +923,6 @@ impl<'a> WordIterator<'a> {
     }
 }
 
-fn filter_empty<'a>(input : &'a str) -> Option<&'a str> {
-    if input == "" { None } else { Some(input) }
-}
-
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks! We're getting closer to completion -- just a handful of little issues like this that need to be ironed out.

PR #351, opened 2017-06-28
Introduce a match-case construct to Ion's grammar

### LINKED ISSUE #273: Implement Match / Switch / Case Flow Keyword
Implementing this will require a bit of studying of the existing **FlowLogic** methods. Shouldn't be too hard though. The hardest part should be determining the syntax that we want to go with.

```ion
match $(getconf LONG_BIT)
    case "32"
        echo "32-bit OS found"
        end
    case "64"
        echo "64-bit OS found"
        end
    case _
        echo "Unsupported OS architecture"
        end
end
```

We could introduce two new statement types: Match and MatchArm. The statement splitter should be able to split upon detecting a complete match arm.

huntergoldstein:
**Problem**: In order for a user to examine an expression and act based on its value, one would have to write a set of nested if statements such as:
```
let v = ...
if test v -eq "foo"
...
else if test v -eq "bar"
...
else
...
end
```

**Solution**: Introduce a match construct that works like a case-switch statement in Bash or some C-like language:
```
match ...
  case foo; ... ; end
  case bar; ... ; end
  case _ ; ... ; end
end
```

- Adds two new statement variants: `Match` and `Case` as well as a `Case` struct
- Update `FlowLogic` to require an `execute_match` function and add associated logic (collecting cases, ensuring that a `Match` block is complete / incomplete).

**TODOs**: 
- [x] Need to add unit and integration tests; @mmstick is there a good place to add these?
- [ ] Syntax extensions (if needed)
  - [ ] Allow variable binding to default case: `case _ as foo; echo $foo ; end`
  - [x] Alternation cases, so something like: `case "bar" | "baz" | ...`
  - [x] Non constant cases? `case $(echo file | ... | ...)`

**Fixes**: Closes #273 

**State**: WIP: see TODOs, though the syntax extensions can be moved into another issue if need be, this is a pretty hefty PR.


src/shell/flow_control.rs
```diff
@@ -76,6 +107,75 @@ pub struct Function {
...
+    where I : Iterator<Item=Statement>
+{
+    while let Some(statement) = iterator.next() {
+        match statement {
+            Statement::Case(case) => {
+                *level += 1;
+                if *level == 2 {
+                    // When the control flow level equals two, this means we are inside the
+                    // body of the match statement and should treat this as the new case of _this_
+                    // match. Otherwise we will
```
mmstick: Incomplete comment here.

mmstick:
The examples directory contains all of our tests for this kind of functionality.

huntergoldstein:
I've updated the `case` branches to be arbitrary expressions. The matched value and each branch will expand to an array and "matching" means there's an intersection between the arrays. That behavior is certainly up for debate. It works as it provides a simple alternation syntax:
```
match $filetype
   case [image/png  image/jpeg]; ... ; end
   ...
end
```
But it also creates strange situations like:
```
match [1 2 3]
   case 1; ...; end
   ...
end
```
An alternative is treating the value we are matching against as a string no matter what, so arrays will get `join(" ")`ed. 

I'm satisfied with this PR otherwise.

mmstick:
It does seem that we will need to brainstorm more sophisticated matching mechanisms to handle all of the possible use cases for matching on strings and arrays. My suggestion would be to offer more than just the base `case` keyword. We could have multiple different types of `case` keywords, such as `case intersects/!intersects`, `case contains/!contains`, `case if`, `case is`.

```ion
match @array
    case contains 1; ...; end
    case !contains 2 && contains 3; ...; end
    ...
end

match $filetype
    case is "image/png" || is "image/jpeg"; ...; end
    case intersects [image/png image/jpeg]; ...; end
    ...
end
```

mmstick:
Merging this as is for now. Thanks for the work!

PR #353, opened 2017-06-28
Fixed completing executables

### LINKED ISSUE #352: Completion for executables is broken
**Reproduction**: Attempt to autocomplete an executable

**Expected behavior**: 
```
$ ba<tab>
bacman     badblocks  baobab     base.py    base32     base64     basename   
bash       bashbug
```
**Actual behavior**: 
```
$ ba<tab>
# nothing happens
```
**Misc**: 
I have confirmed that this bug was introduced in [this commit](https://github.com/redox-os/ion/commit/4cc422eb383653333a744b0266dc3f35ebd36c7b), of which I am the author.

huntergoldstein:
The problem was that for any filename "foo", `readln` would check to see if either it exists, or if the parent exists. Which means if you're trying to complete _any_ executable, it thinks its a file because the "parent" is the current directory.

Closes #352.

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -78,6 +78,9 @@ impl<'a> Shell<'a> {
         }
     }
 
+    /// Infer if the given filename is actually a partial filename by determining if
+    /// the file exsits in the current directory, or if the parent of the file exists in the
+    /// current directory, and is not the directory itself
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks for fixing it!

PR #354, opened 2017-06-28
Additional fix for #352

huntergoldstein:
Due to the condition that the parent directory cannot be the current directory, simple commands like `rm file_in_this_dir.txt` were broken.


huntergoldstein:
Hold off on merging this for now: I'm taking the time to ensure that the logic in `readln` is exactly what we want.

huntergoldstein:
Good to go now; expanded the logic and made it much more readable.

Issue #29, opened 2016-01-18
Ion needs job control


skylerberg:
We need to allow interactive job control. We will need the following commands:
- [x] bg
- [ ] fg (Partially Implemented)
- [x] jobs
- [x] suspend
- [x] disown
- [x] wait
- [x] process groups

skylerberg:
https://github.com/redox-os/ion/pull/86

skylerberg:
We can now run a task in the background. There is still much work to be done.

skylerberg:
Since  the backgrounding was not useful in it's current state and I haven't taken the time to make it work properly with piping, I am removing it. This is not really lost work because I am going to add it back in for pipelines sometime soon. I am also leaving it in the grammar for now. For reference, here is the code I am taking out: https://github.com/redox-os/ion/blob/4a8cd17ec85d1e2a980987c2dda7635291f77653/src/main.rs#L228.

mmstick:
This work should be very simple to implement now. We just need a means of storing job thread handles in the shell, and then implement the commands to send signals to these running jobs.

mmstick:
SIGINT and SIGTERM are implemented, and process IDs are now stored within the shell.

mmstick:
SIGTSTOP is implemented.

mmstick:
We now have `fg`, `bg`, `wait`, and `jobs`.

mmstick:
I will note, however, that, I observed some strange behavior with the difference between the reported PID by Rust, and the actual PID found in `top`. I had to apply an offset based on how a process is getting sent to the background. A command like `command args... &` would get a PID that's off by 2 of the actual PID, whereas `command args...` would have a PID that's off by 1. If anyone can figure out why that is, I'd like to know.

myfreeweb:
I'm happy that ion is finally getting job control… however it's not really there yet.

First, don't hardcode SIGTSTP as 20. It's not 20 on FreeBSD, for example. Use the constant from `libc`/`nix`.

Second, seems like the entire TSTP handling thing is wrong :( That's not how other shells do it.

Currently just suspending vim inside ion that was started from zsh returns… back to the zsh.

[...]

PR #362, opened 2017-07-02
Job control: libc constants, pid fix

myfreeweb:
This PR removes hardcoded signal numbers and adds `-1` to the PID when sending SIGCONT. #29

Tested on FreeBSD, Arch Linux and the Windows Subsystem for Linux. Redox might be even more broken though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ Below is an overview of features that Ion has either already implemented, or aim
 - [x] Background Jobs
 - [ ] Piping Functions
 - [ ] Redirecting Functions
-- [ ] Background Job Control
+- [x] Background Job Control
 - [ ] XDG App Dirs
 - [ ] Plugins Support
     - [ ] Builtins
--- a/src/builtins/job_control.rs
+++ b/src/builtins/job_control.rs
@@ -5,7 +5,7 @@ use std::io::{stderr, Write};
 use std::thread::sleep;
 use std::time::Duration;
 #[cfg(not(target_os = "redox"))] use nix::sys::signal::{self, Signal};
-#[cfg(not(target_os = "redox"))] use libc::pid_t;
+#[cfg(not(target_os = "redox"))] use libc::{self, pid_t};
 
 /// Display a list of all jobs running in the background.
 pub fn jobs(shell: &mut Shell) {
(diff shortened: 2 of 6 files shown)


myfreeweb:
Wait, I'm actually not sure about `-1`, maybe it should be added to all `signal::kill` calls…

myfreeweb:
Yeah, just tried printing the kill result for background processes in the `exit_builtin` — it's `Ok` only after adding `-1`, and `Err(Sys(ESRCH))` without the fix, on both FreeBSD and Linux.

mmstick:
PID offsets are set [here](https://github.com/redox-os/ion/blob/master/src/shell/job_control.rs#L114). For whatever reason, I've had to employ a different offset for different scenarios to get everything working on my end. When a command is executed as `command args...`, it's off by 1 from the PID reported by `top`, whereas if the command is executed with `command args... &`, it's off by 2.

myfreeweb:
Oh right. Maybe your offsets are off by one then — it seems to actually be 0 and 1 instead of 1 and 2…

mmstick:
I'd like to find out why Rust reports a different PID for the child process than what I observe from `top`.

myfreeweb:
It actually doesn't?? I just checked on both FreeBSD and Linux, with the offset removed, it is not different from `top` and `htop`!

mmstick:
Then I wonder what's wrong with my system (Running Solus w/ Linux 4.9). Without applying the offsets, everything's off by 1 or 2. I'll try rebooting.

mmstick:
Strangely, the offset is still required even after reboot.

myfreeweb:
I don't think rebooting can help :D

What version of rustc are you running?

mmstick:
It's one of the 1.20 nightly builds. I'm not at home at the moment, so I can't give a specific version.

myfreeweb:
Just tried a CentOS 7 server, no offset required either.

mmstick:
Going to try it out on my Arch desktop.

mmstick:
My Arch desktop requires the same PID offsets to function. For reference, I'm testing with the `stress` command on Arch (`stress-ng` on Solus).

mmstick:
Also interesting to note is that `stress` itself reports it's own PID when it launches, which is identical to the PID Rust gives without the offset, but the actual PID reported in `top` is always offset by 1 or 2, and sending signals to the PID reported by Rust and the stress command has no effect.

mmstick:
I'll merge this PR anyway. I'll have to investigate further into this oddness.

myfreeweb:
> stress itself reports it's own PID when it launches, which is identical to the PID Rust gives without the offset

That must mean that it forks itself once or twice and the original process dies! Like I said, that's why process groups are necessary.

I tested with `pacman`, `dd` and `nvim`, they don't do that.

Issue #326, opened 2017-06-22
Store Configurations In the Correct XDG App Dir


mmstick:
Currently, Ion just slops the init file in $HOME, but we should use the correct directory according to XDG specifications. The [app_dirs](https://crates.io/crates/app_dirs) crate is just the thing that would make this simple to implement.

huntergoldstein:
#363 closes this correct?

mmstick:
Yes

PR #363, opened 2017-07-04
Store configurations in the correct XDG App Dir

clippix:
Read init file from the XDG App Dir using the app_dirs crate.
The init file is now named initrc instead of .ionrc.
---

Fix Issue #326 
I wasn't sure how to format some of these changes (like the author name or the struct declaration inside a function call), please feel free to modify it.

Moreover, it looks like some others things could be moved into the app dir (history file)?

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -27,6 +27,7 @@ use std::sync::{Arc, Mutex};
 use std::sync::mpsc::Receiver;
 use smallvec::SmallVec;
 
+use app_dirs::{AppDataType, AppInfo, app_root};
 use builtins::*;
 use fnv::FnvHashMap;
 use liner::{Context, CursorPosition, Event, EventKind, BasicCompleter};
@@ -363,14 +364,15 @@ impl<'a> Shell<'a> {
 
     /// Evaluates the source init file in the user's home directory.
     pub fn evaluate_init_file(&mut self) {
-        env::home_dir().map_or_else(|| {
-            let stderr = io::stderr();
-            let mut stderr = stderr.lock();
-            let _ = stderr.write_all(b"ion: could not get home directory");
-        }, |mut source_file| {
-            source_file.push(".ionrc");
-            self.execute_script(&source_file);
-        });
+        match app_root(AppDataType::UserConfig, &AppInfo{ name: "ion", author: "Redox OS Developers" }) {
+            Ok(mut source_file) => {
+                source_file.push("initrc");
+                self.execute_script(&source_file);
+            },
+            Err(_) => { // The error returned could be used for a possibly more descriptive error
+                eprintln!("ion: could not get home directory");
+            }
+        }
     }
 
     pub fn prompt(&self) -> String {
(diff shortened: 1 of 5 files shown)


src/shell/mod.rs
```diff
@@ -363,14 +364,17 @@ impl<'a> Shell<'a> {
...
     pub fn evaluate_init_file(&mut self) {
-        env::home_dir().map_or_else(|| {
-            let stderr = io::stderr();
-            let mut stderr = stderr.lock();
-            let _ = stderr.write_all(b"ion: could not get home directory");
-        }, |mut source_file| {
-            source_file.push(".ionrc");
-            self.execute_script(&source_file);
-        });
+        match get_app_root(AppDataType::UserConfig, &AppInfo{ name : "ion", author : "mmstick" }) {
```
mmstick: How about making the author as `Redox OS Developers`?

mmstick:
We could also move the history file as well, by default.

src/shell/mod.rs
```diff
@@ -363,14 +364,17 @@ impl<'a> Shell<'a> {
...
+            Ok(mut source_file) => {
+                source_file.push("initrc");
+                self.execute_script(&source_file);
+            },
+            Err(_) => { // The error returned could be used for a possibly more descriptive error
+                let stderr = io::stderr();
+                let mut stderr = stderr.lock();
+                let _ = stderr.write_all(b"ion: could not get home directory");
```
mmstick: I think it's okay for us to use `eprintln!()` now:

```rust
eprintln!("ion: could not get home directory");
```

huntergoldstein: `eprintln` is in the `1.19.0-beta.2` so presumably it will be part of the official `1.19.0` release.

mmstick: And Ion currently requires 1.19 as a minimum version at the moment.

clippix:
Thanks for the suggestions, I will submit an update.

clippix:
Is there a preferred way to submit updated PRs?
* Should I commit over the current one?
* Or should I add changes and commit them based upon the existing one?
(Sorry, I'm more used to the "submit patches by mail" scheme).

Moreover, should ion create the needed directories (~/.config/ion and ~/.local/share/ion for example) when they don't exist yet? (If not and HISTORY_FILE_ENABLE is set to 1, it complains that the history cannot be written, even by the winners)

mmstick:
The usual git workflow is:

```sh
git add *
git commit -m "Title of Commit

Description of commit"
git push local branch_name
```

You could also perform `git commit --amend" to overwrite a commit, and use `--force` when pushing the overwrite. In addition, if you push to a branch with the branch name prefixed with `:`, it will delete that branch.

When you push any changes to the branch that you are trying to merge here, this page will automatically update, and Travis will run a new test on your new commit.

mmstick:
Also, yes, it should create the directories if they do not exist. The history file should be enabled by default, so that's a change to make within Ion itself.

clippix:
Okay, I didn't read carefully the part about the commits being collapsed together, so the enabling of history feels a bit off topic, sorry.

mmstick:
Thanks for the work! It's merged now.

Issue #331, opened 2017-06-22
Implement &| (Stdout + Stderr Piping)


mmstick:
At the moment, `&|` is equivalent to `^|`, in that it only maps stderr, instead of both stdout and stderr. This will be important to resolve before the 1.1 release.

huntergoldstein:
This looks like it works as expected (aside from a strange extra newline)
[![asciicast](https://asciinema.org/a/N0xTAlfQe2bl3OaRuiB7bgQmu.png)](https://asciinema.org/a/N0xTAlfQe2bl3OaRuiB7bgQmu)

mmstick:
I believe it's the piping (&|) that specifically doesn't work, as it just copies the behavior of piping stderr (^|)

huntergoldstein:
Confirmed: I don't think `parser::pipelines` recognizes it as it looks like it parses `foo &| bar` as "fork `foo` to the background, run `| bar`."

mmstick:
The main issue is that there's no way to specify two file descriptors for the stdin of the next command in line. You may only specify one file descriptor at a time, which means that you can only either listen to stderr, or stdout. The fix would entail somehow creating a new, internal file descriptor in memory as a buffer, and supplying that temporary file descriptor for the stdin of the next command.

huntergoldstein:
Yup I realize that now; I thought I could get away with just `dup` trickery.

mmstick:
I was thinking that we could perhaps use [memfd_create](https://docs.rs/nix/0.8.1/nix/sys/memfd/fn.memfd_create.html). It will create a memory region and return a file descriptor to it, which we could map the stdout and stderr to, and then the stdin of the next process could be that region.

huntergoldstein:
How about for Redox? Part of the hope of just using `dup` was that there are `dupN` syscalls for Redox.

mmstick:
@huntergoldstein Do you think you could implement it with the `dup` syscalls?

huntergoldstein:
I think what `memfd_create` does can also be done with pipes, which I would assume Redox has _some_ form of. I'll have to play around with it more, I've been looking at updating `statement.rs` and `pipeline.rs`.

huntergoldstein:
Taking a closer look now; I tried to rewrite `statements.rs` and `pipelines.rs` over the weekend, I'll open up an issue with my findings.

huntergoldstein:
[...]

PR #365, opened 2017-07-05
Fix redox build

huntergoldstein:
Redox build was breaking for a number of reasons:

- Apparently the crate for Redox syscalls is just `syscall`, which is news to me
- `libc` is only imported for Unix and non-redox, which means any case where we refer to a `libc` constant needs to be behind a `cfg`
- The Redox version of `watch_pid` had a different number of arguments


huntergoldstein:
Also this effectively blocks #331 as I want to test on redox.

mmstick:
Thanks!

mortonar:
Thanks! Was running into these issues as well.

PR #368, opened 2017-07-07
Implement `&|` stdout and stderr piping

### LINKED ISSUE #331: Implement &| (Stdout + Stderr Piping)

huntergoldstein:
- `&|` is now part of the grammar as the stdout and stderr piping operator
- Redirecting stdout and stderr with `&|` is now supported on *nix and Redox

**Drawbacks**:
- The big `crossplat` module is a little messy. Perhaps moving all of the platform specific code to a separate file altogether would be better. It seems like Rust's pattern for this is something like:
```
#[cfg(windows), path = 'for_windows.rs']
mod some_functionality;

#[cfg(unix), path = 'for_unix.rs']
mod some_functionality;

...
```

**Fixes**: Closes #331 

**State**: Ready to go, pending review.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,17 +3,18 @@ name = "ion-shell"
 version = "1.0.3"
 dependencies = [
  "ansi_term 0.9.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "app_dirs 1.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "bitflags 0.9.1 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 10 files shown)


mmstick:
Thanks for the work! I'm about to land some big changes myself to the pipe logic -- process groups, signal masking, foreground management. Albeit, I've yet to figure out why my `fg` implementation isn't working. Perhaps has something to do with performing `waitpid` in a background thread, and then the foreground process needing to take control of and perform `waitpid` itself.

mmstick:
And merged

PR #372, opened 2017-07-08
Changed around some small things so calc gets pulled from it's own crate

mgmoens:
I think the code speaks for itself I guess? Don't know if this was reserved, if it's not the right moment please just discard this. Also, is it me or has Cargo.lock been removed from the .gitignore?


mmstick:
Thanks! Binary projects shouldn't have their `Cargo.lock` file ignored.

PR #373, opened 2017-07-08
Fix autocomplete for partially escaped filenames

### LINKED ISSUE #348: double escaping
touch '(x)'
    rm \(<Tab> # rm \\(x\)
    ion: syntax error: '(' at position 6 is out of place

huntergoldstein:
**Fixes**: Closes #348.

- Add `unescape` which maps from escaped strings to their raw counterparts
- Add custom logic for determining word boundaries for `liner` completions; this is based off the default completer with some adjustments for backslash.

**Drawbacks**: 
- I can't remember who, but someone mentioned that we ought to have some central list of what should get escaped / un-escaped; this builds on the problem without fixing it. Additionally this does not allow something like:
```
$ touch "foo(bar)"
$ rm "foo(<Tab>
```
to autocomplete `"foo(` to `"foo(bar)"`.

**State**: Ready to go!

### DIFF
--- a/src/shell/completer.rs
+++ b/src/shell/completer.rs
@@ -86,7 +86,7 @@ impl Completer for IonFileCompleter {
             }
         }
 
-        self.inner.completions(start).iter().map(|x| escape(x.as_str())).collect()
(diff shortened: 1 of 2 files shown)


mmstick:
Thanks!

Issue #132, opened 2016-03-07
Add heredocs


skylerberg:
```
cat <<EOF
some
stuff
EOF
```

mgmoens:
Can someone tell me where |, > etc. are parsed? I'd like to try this if I know where those special characters get parsed. I've been looking around in the shell folder, but I think I'm looking to deep in the rabbit hole.

mmstick:
They are parsed [here](https://github.com/redox-os/ion/blob/master/src/parser/pipelines.rs#L198) and [here](https://github.com/redox-os/ion/blob/master/src/parser/pipelines.rs#L172) in the pipelines module. At some point, I'd like to rewrite this module. Haven't gotten around to it though.

mmstick:
I'll have heredocs implemented in a few minutes.

PR #382, opened 2017-07-10
Rewrite `pipelines::collect` to be a recursive descent parser

huntergoldstein:
**Problem**: The previous version of `pipelines::collect` was a dragon's nest of bitflags and `break <label>`. It was difficult to extend for features such as #132 and #133.

**Solution**: `pipelines::collect` has been replaced by `pipelines::Collector`, which encapsulates the logic of `pipelines::collect` in a recursive descent parser. Additionally, instead of taking a mutable reference to an `Option<&str>`, `pipelines::Collector::run` returns a `Result<...>`.

- Removed `pipelines::collect`
- Added `pipelines::Collector` and replaced instances of `pipelines::collect` with the logical equivalent for `pipelines::Collector`.

**Drawbacks**:
- The error messages are still of questionable quality. The main problem is being constrained by `peg`: the error messages have to be `&str`s according to the documentation, which is just mind boggling.
- There's still a good amount of complexity, cyclomatic complexity warning notwithstanding

**TODOs**:
- [x] Add tests for obscure redirects like `&>>` and `^>>`

**State**: WIP just so that it has more time for review. Also see TODOs.

**Other**: In addition to the mentioned issues I'm trying to figure out what to tackle next. Replacing PEG is a must, but there's also `statements.rs` which needs a similar makeover.


mmstick:
Something we might also want to do is have the ability to retain access to the original string that made up that pipeline, so that if a pipeline happens to be executed within a fork in the background, we can supply that string as the command when printing jobs via `jobs`.

huntergoldstein:
So the `Pipeline` would keep track of the original string? Makes sense to me.

huntergoldstein:
Would we also want to expand the original command? For example if I run the pipeline `cat $FOO` where `FOO = bar.baz`, I would imagine we'd want to display `cat bar.baz` as the running command, as `cat $FOO` is ambiguous. At the same time that might get unruly if you have a command like:
```
parallel sed -i 's/BadName/GoodName/g' {} ::: $(ag -l BadName)
```

mmstick:
I think it should be expanded, so the string should be created when the pipeline is being expanded, but perhaps truncated if the string is too long.

huntergoldstein:
There's a small problem in that `WordIterator` might break in unexpected ways if we run it on the full string: let me see what I can't do.

huntergoldstein:
So I've written a custom implementation of `Display` for `Pipeline` which "recreates" the pipeline. It forces an order, `JOBS [REDIR_IN] [REDIR_OUT]`, but this should serve your purposes well @mmstick. This can be called after `Pipeline::expand` and you'll get the expanded form of the pipeline.

mmstick:
That will be perfect, especially given that the command name will always be requested after the pipeline is expanded (during pipeline execution).

huntergoldstein:
I'm good to have this merged in then

mmstick:
Thanks for the work!

mmstick:
Job commands are now displayed in the output of `jobs` thanks to this PR: https://github.com/redox-os/ion/commit/c6f90de62a54868cbb2ec34b060cb5af6e499716

mmstick:
This also made it possible to implement `set -x`: https://github.com/redox-os/ion/commit/8f3df249de8b7c410c83e1574f6e1de7756212f2

PR #403, opened 2017-07-12
Fix build by changing Option field

clippix:
Introduced by commit ea4b98c94f21, an Option modifier on Context field in the
Shell struct.


huntergoldstein:
@mmstick is there a reason that `context` was changed to an optional?

mmstick:
Was planning to have it be set to `None` for script executions, but forgot that I left that in there.

PR #404, opened 2017-07-12
Add `array!` macro for creating inline `Array`s

huntergoldstein:
**Problem**: Writing `Array::from_vec(vec![...])` is verbose and performs more computation than needed.

**Solution**: Add an `array!` macro that replaces any instance of `Array::from_vec(vec![...])` with `array![...]`. Additionally it calls `into` on its elements, so instead of having to write:
```
Array::from_vec(vec!["foo".into(), "bar".into()])
```
it will be sufficient to write
```
array!["foo", "bar"]
```

**Drawbacks**: This creates an extra level of indirection: it is unclear what type the `array!` macro refers to. However I don't think it is any more indirection from before, where the `Array` type is just a `SmallVec` alias.

### DIFF
--- a/src/types.rs
+++ b/src/types.rs
@@ -8,7 +8,7 @@ pub type Value = String;
 pub type VariableContext = FnvHashMap<Identifier, Value>;
 pub type ArrayVariableContext = FnvHashMap<Identifier, Array>;
 
-/// Construct a new Array containing the given artuments
+/// Construct a new Array containing the given arguments
 ///
 /// `array!` acts like the standard library's `vec!` macro, and can be thought
 /// of as a shorthand for:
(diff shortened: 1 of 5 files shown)


mmstick:
I would as some documentation to the macro. Then it will be easy to figure out the what and the why of the macro.

src/types.rs
```diff
@@ -7,3 +7,29 @@ pub type Identifier = SmallString;
 pub type Value = String;
 pub type VariableContext = FnvHashMap<Identifier, Value>;
 pub type ArrayVariableContext = FnvHashMap<Identifier, Array>;
+
+/// Construct a new Array containing the given artuments
```
mmstick: Found spelling mistake here: `artuments`

mmstick:
Thanks!

PR #405, opened 2017-07-12
[WIP] Add support for herestrings

### LINKED ISSUE #133: Add herestrings
```
cat <<<this string is the input
```

huntergoldstein:
- Modified `Pipeline::stdin`'s type to be a union of either a filename or a string input
- Added support in `pipelines::Collector` for recognizing herestrings
- Added required functionality in `Shell::execute_pipeline`

**TODOs**:
- [ ] **Outstanding Bug**: running `tr '[a-z]' '[A-Z]' <<< "foo"` produces no output
- [x] Add doc comments for the `peg::Input` enum
- [x] Add regression tests in `pipelines.rs` and integration tests in `examples/` for herestrings

**Fixes**: Closes #133

**State**: WIP; see TODOs. 

**Other**: This PR ended up being slightly different than #200; instead of holding onto the string and writing to `stdin` once the command had been spawned, this takes a cue from the implementation of `&|`: write the herestring to a pipe, close the writer end, and pass the reader end to the first command in the pipeline.

### DIFF
--- a/src/parser/peg.rs
+++ b/src/parser/peg.rs
@@ -19,14 +19,20 @@ pub struct Redirection {
 }
 
 #[derive(Debug, PartialEq, Clone)]
+pub enum Input {
+    File(String),
+    HereString(String),
+}
+
+#[derive(Debug, PartialEq, Clone)]
 pub struct Pipeline {
     pub jobs:   Vec<Job>,
     pub stdout: Option<Redirection>,
-    pub stdin:  Option<Redirection>,
+    pub stdin:  Option<Input>,
 }
 
 impl Pipeline {
-    pub fn new(jobs: Vec<Job>, stdin: Option<Redirection>, stdout: Option<Redirection>) -> Self {
+    pub fn new(jobs: Vec<Job>, stdin: Option<Input>, stdout: Option<Redirection>) -> Self {
         Pipeline { jobs, stdin, stdout }
     }
 
@@ -36,9 +42,17 @@ impl Pipeline {
             job.expand(&expanders);
         }
 
-        if let Some(stdin) = self.stdin.iter_mut().next() {
-            stdin.file = expand_string(stdin.file.as_str(), &expanders, false).join(" ");
-        }
+        let stdin = match self.stdin {
+            Some(Input::File(ref s)) => {
+                Some(Input::File(expand_string(s, &expanders, false).join(" ")))
+            },
+            Some(Input::HereString(ref s)) => {
+                Some(Input::HereString(expand_string(s, &expanders, false).join(" ")))
+            },
+            None => None
+        };
+
+        self.stdin = stdin;
 
         if let Some(stdout) = self.stdout.iter_mut().next() {
             stdout.file = expand_string(stdout.file.as_str(), &expanders, false).join(" ");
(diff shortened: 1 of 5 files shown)


huntergoldstein:
- I think the way the prompt is displayed is really messing with the output of some of these programs being displayed:
[asciicast](https://asciinema.org/a/bEGKkmKSnevCw4j9MFynCBRnZ)
- Is there a known reason why using a subshell expansion causes the newlines to be stripped?

mmstick:
Newlines in command substitutions are handled the same way as in Bash -- they are converted into spaces unless you double quote the substitution.

huntergoldstein:
I found [this piece of documentation](http://tldp.org/LDP/abs/html/commandsub.html) which says that this will happen when echoing an unquoted variable created with newline, but the following happens on `bash`:
```
$[hunter@ss-arch ~]$ tr '[a-z]' '[A-Z]' <<< $(ls)
BACKUPS
BUILDS
CMR
DOCUMENTS
DOWNLOADS
DROPBOX
MOUNTED
MUSIC
PICTURES
PLAYONLINUX.D
SECONDARY
SUPERPROJECT
VIDEOS
VIRTUAL MACHINES
```
I get the feeling that behavior mentioned in the documentation happens implicitly: the newlines cause the args to get split but `echo` just prints each one out after the other.

mmstick:
We could reverse the quoting rules in herestrings as it's done for for loops.

huntergoldstein:
See #410; I'll just assume the behavior isn't changing for this PR.

src/parser/peg.rs
```diff
@@ -36,9 +45,17 @@ impl Pipeline {
...
 
-        if let Some(stdin) = self.stdin.iter_mut().next() {
-            stdin.file = expand_string(stdin.file.as_str(), &expanders, false).join(" ");
-        }
+        let stdin = match self.stdin {
+            Some(Input::File(ref s)) => {
+                Some(Input::File(expand_string(s, &expanders, false).join(" ")))
+            },
+            Some(Input::HereString(ref s)) => {
+                Some(Input::HereString(expand_string(s, &expanders, false).join(" ")))
```
mmstick: If you change the `false` to `true` here, the newlines won't be replaced with spaces.

src/shell/pipe.rs
```diff
@@ -159,15 +187,30 @@ impl<'a> PipelineExecution for Shell<'a> {
...
+                        },
+                        Err(e) => {
+                            eprintln!("ion: failed to redirect '{}' into stdin: {}", filename, e);
+                        }
+                    }
+                }
+            },
+            Some(Input::HereString(ref string)) => {
+                if let Some(command) = piped_commands.first_mut() {
+                    match unsafe { crossplat::stdin_of(string) } {
```
mmstick: Adding the following fixes the issue with the "foo" argument:

```rust
let result = if string.ends_with('\n') {
    unsafe { crossplat::stdin_of(string) }
} else {
    let string = [string, "\n"].concat();
    unsafe { crossplat::stdin_of(&string) }
};
```

Doing a match on that `result` will get you what you want.

mmstick:
Is this ready to merge now?

huntergoldstein:
Yup; good to go

mmstick:
Thanks!

PR #407, opened 2017-07-13
Tildes are now part of normal tokens

### LINKED ISSUE #406: [DANGER] Tidle is expanded into separate word when used with globbing
**Reproduction**: Attempt to run a command that includes both globing and a tilde.

**Expected behavior**:
```
$ ls ~/Downloads/
cd61.iso
some.program.tar.gz
$ rm ~/Downloads/*
$ ls ~/Downloads
< nothing >
$
```

**Actual behavior**:
```
$ ls ~/Downloads/
cd61.iso
some.program.tar.gz
$ rm ~/Downloads/*
rm: cannot remove '/home/hunter': Is a directory
```

**Build information**: 
- `git rev-parse HEAD`: 244811eb8d58c07dd546cadc5abd0cc506f6e6bc

**Misc**: This is concerning and I don't think the fix is trivial. I'm going to try and get this done right now but I'm gone for the next few days and this seems urgent.

huntergoldstein:
**Fixes**: Closes #406.

**Summary**: Instead of parsing tildes separately from normal tokes they are now included inside normal tokens, but the `tilde` flag is set so when they are expanded the `ExpansionFunction::tilde` function is called on the string.

**Drawbacks**: 
- Very untested (but all the tests pass, and the examples)
- I'm not sure this is the optimal solution, but it works
- This is dependent on `expand_func.tilde` expecting a full path that is prefixed with a tilde; I'm not sure if we're ok with that.

### DIFF
--- a/src/parser/shell_expand/mod.rs
+++ b/src/parser/shell_expand/mod.rs
@@ -159,6 +159,31 @@ pub fn expand_tokens<'a>(token_buffer: &[WordToken], expand_func: &'a ExpanderFu
     let mut output = String::new();
     let mut expanded_words = Array::new();
     let mut is_glob = false;
+
+    macro_rules! expand {
+        ($text:expr, $do_glob:expr, $tilde:expr) => {{
+            let expanded: String = if $tilde {
+                match (expand_func.tilde)($text) {
+                    Some(s) => s,
+                    None => $text.into()
+                }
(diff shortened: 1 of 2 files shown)


mmstick:
We can find better solutions in the future if need be. I think this is perfectly fine for now. Thanks!

PR #427, opened 2017-07-17
bugfix: echo was treating anything preceded by - as an option flag

drosseau:
`echo` was treating everything preceded by - as an options flag so something like `echo -1` wouldn't output anything.

This change would make it act more like the usual `echo`, which would even take something like `echo -n1` and output `-n1` while `echo -n -1` would recognize the `-n` flag and output `-1` without a newline.


mmstick:
Thanks! Would you be interested in submitting more contributions? There's a lot of easy jobs in the builtin department of the shell that we need implemented, and I have a quick PR turn around time.

drosseau:
Yea, I actually found this bug while implementing `not`. I'm trying to work on "low hanging fruit" issues for now while I get familiar with the code.

PR #428, opened 2017-07-17
Jobs no longer use a separate list of builtins to determine the command type

drosseau:
Before when determining how to create the command the code would look at a hand maintained list which had to be updated whenever a new builtin was added. It looked like that was fairly out of date and caused some odd behavior. For example `true` and `false` weren't in the list so `true && false` wouldn't be using the builtins, but just `true` or `false` alone would be. Since the shell has the builtins hash I figured we can just check if it is in there or not by checking the keys and then call the appropriate method to create the command.


mmstick:
Thanks!

PR #429, opened 2017-07-17
Conditionals && and || weren't working as expected

drosseau:
If you ran `false && echo and || echo or` you should have gotten "or" echoed, but instead nothing was coming out. It was caused by advancing these iterators. It looks like those lines were added [here](https://github.com/redox-os/ion/commit/9d5cb19d26d0744c9de48983d981b4d0da4984b7#diff-cc20bc374fba8d873a99948f7710250eL127), but I'm not exactly sure why. It references `&|`, but that seems to work fine without them.

Also added that examples script to help make sure they're still working in the future.

### DIFF
--- a/examples/conditionals.ion
+++ b/examples/conditionals.ion
@@ -1,0 +1,9 @@
+true && echo and || echo or
+false && echo and || echo or
+false || echo false
+true && echo true
+true || echo cant get here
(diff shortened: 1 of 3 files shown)


mmstick:
Thanks!

PR #430, opened 2017-07-18
Read init file only when using interactive shell

clippix:
Execute the initrc file only when using ion in interactive mode as it probably makes less sense in the other modes.

Moreover, it fixes modification of the context done in the initrc file which weren't taken into account as the context wasn't initialized at the time of the script execution (e.g.: `set -o vi` in the initrc didn't set the vi editing mode in interactive sessions).


mmstick:
Thanks!

Issue #419, opened 2017-07-16
Fix `echo hello#world`


mmstick:
The output used to be `hello#world`, but it seems the parsing has changed so that the `#world` is cut off. This is more than likely related to the `StatementSplitter`. We should implement tests to ensure that it doesn't happen again in the future.

PR #432, opened 2017-07-18
Fix issue #419 `echo hello#world`

drosseau:
This should fix issue #419, but I'm not sure if it is exactly how we want to do it. With this change any `#` not at the start of the line (`self.read == 1`) or preceded by an ascii whitespace character (`b' ' | b'\t'`) will be considered part of the token it is inside.


mmstick:
Thanks. This is how it was working before, and is the standard behavior in other shells.

PR #433, opened 2017-07-19
Add support for inline expressions in methods

### LINKED ISSUE #389: Implement Inline Methods
I've decided that we will pursue implementing this on top of our existing method syntax. The rules are simple, but might require some complex brain work to get the parser ready and capable of parsing the additional rule. The purpose of inline methods is to prevent the need to round trip results to variables and increasing the number of lines of code to parse and execute, costing time and energy, of which we have limited amounts of both.

Basically, it works as follows: if the input parameter in a method begins with `$`, `@`, `[`, `"`, or `'`, then that parameter is an expression and should be parsed as such. Otherwise, treat the parameter as the preferred variable type.

# As Expression

```ion
let lines = @lines($read("filename"))
let lines = @lines($(cmd args...))
let lines = @lines($string_var)
```

# As Variable

```ion
let lines = @lines(string_var)
```

huntergoldstein:
- If the "variable" for a string or method looks like an expression (starts with `@`, `$`, etc.), it is expanded as if it were an expression, otherwise it is treated as a variable

**Drawbacks**: I feel like this PR has exacerbated some issues with how methods are parsed, for example:
```
$ let foo = bar
$ echo @len( @bytes(foo))
0
$ echo $len( foo)
<no output>
```
I'd like to make a sweep over how methods are parsed and do things like consolidating the logic for all the string methods, making it easier to pass in arbitrary values to methods, making `StringMethod`s work exactly like `ArrayMethod`s or vice-versa, etc. Also fixing the above.

**Fixes**: Closes #389

**State**: Good to go aside from **Drawbacks**

### DIFF
--- a/src/parser/shell_expand/mod.rs
+++ b/src/parser/shell_expand/mod.rs
@@ -18,6 +18,17 @@ use shell::variables::Variables;
 use std::io::{self, Write};
 use types::*;
 
+/// Determines whether an input string is expression-like as compared to a
+/// bare word. For example, strings starting with '"', '\'', '@', or '$' are
+/// all expressions
+pub fn is_expression(s: &str) -> bool {
+    s.starts_with('@') ||
+    s.starts_with('[') ||
+    s.starts_with('$') ||
+    s.starts_with('"') ||
+    s.starts_with('\'')
+}
+
 pub struct ExpanderFunctions<'f> {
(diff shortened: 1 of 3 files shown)


mmstick:
Thanks! This is a really critical piece of functionality to give Ion an edge over other shells.

PR #435, opened 2017-07-19
Miscellaneous method updates / QoL changes

huntergoldstein:
- `WordToken::StringMethod` now has an internal `struct` which manages the logic for string methods
- Changed `slice_string` to `slice` and allowed any input that satisfies `AsRef<str>`

**State**: Good to go!

------

Aside: I'm wondering if `WordToken` could be benefited from being more token-like: more lightweight than they are now. For example, instead of having an `ArrayProcess` that stores a string that will be tokenized later, we can just have an `ArrayProcessBegin` and `ArrayProcessEnd` then tokenize the contents.


mmstick:
Thanks!

PR #438, opened 2017-07-19
Case statements are now implicitly ended

### LINKED ISSUE #414: Make End Keyword In Case Statements Optional
This is what Fish does, as the `case` keyword already denotes that a new case block is being specified, so I'm not sure I see the value in requiring the end keyword to end a case.

huntergoldstein:
- Instead of using `end` to signal the end of a case, case blocks are implicitly ended by either another `case` statement or the end of the `match` block.

**Fixes**: Closes #414 

**State**: Good to go!

**Other**: The description of #414 was "make end optional" but I figure we can just remove it altogether; thoughts?

### DIFF
--- a/examples/match.ion
+++ b/examples/match.ion
@@ -1,7 +1,7 @@
 for i in [1 2 3 4 5 6]
   match $((i % 2))
-    case 0; echo "Even!"; end
-    case 1; echo "Odd!" ; end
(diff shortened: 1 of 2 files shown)


mmstick:
Thanks! I think it's okay to remove the `end` entirely.

PR #440, opened 2017-07-19
Making the builtins a little neater. Closures weren't very friendly for debuggers/backtrace

drosseau:
There are two changes in here.

1) I noticed `unalias` was at some point accidentally defined as a `drop` so we had `drop` down twice.
2) The closure method made it a little more difficult than it needed to be to see what errored with RUST_BACKTRACE=1 and to set breakpoints in a debugger. Now instead of `{{closure}}` you'll get the actual function.

### DIFF
--- a/src/builtins/mod.rs
+++ b/src/builtins/mod.rs
@@ -31,8 +31,11 @@ fn exit_builtin() -> Builtin {
         help: "To exit the curent session",
         main: Box::new(|args: &[&str], shell: &mut Shell| -> i32 {
             let previous_status = shell.previous_status;
-            shell.exit(args.get(1).and_then(|status| status.parse::<i32>().ok())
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks! I'm also wondering if maybe using `OrderMap` instead of `FnvHashMap` would be faster, or a means of lazily-generating the map at compile-time, in order to eliminate the runtime cost of creating the map.

drosseau:
I was looking into the latter and I came across this: https://docs.rs/phf/0.7.21/phf/ but I haven't really looked into how it works.

mmstick:
We would definitely want an ordered map. That might enable us to substitute builtin text at an earlier stage of parsing with the index value that the builtin maps to, to avoid having to perform a builtin lookup repeatedly during script execution.

PR #442, opened 2017-07-20
Some more consistency changes to builtins

drosseau:
We don't need the `Box<Fn(...)>` anymore in the builtins. The only thing this new `Builtin` doesn't allow is closures that capture an environment.

I'm going to start working on changing builtins to accept stdin, stdout, and stderr as parameters now.


mmstick:
Thanks! This a big help to the team!

PR #444, opened 2017-07-21
Added `matches` builtin and fixed grammar issue

drosseau:
Adding matches made me run into an issue with the grammar. Trying to run `matches hello '\w+'` actually started a `match` statement followed by the expression `es hello '\w+'` which was due to having `whitespace?` instead of `whitespace+`. This issue was in a few places. You could declare variables like `letfoo = bar` for instance.


mmstick:
Thanks!

PR #447, opened 2017-07-21
Fix for range syntax bugs

### LINKED ISSUE #445: Brace Ranges Are Invalid
- The exclusive range syntax (`..`) is incorrect. `echo {1..5}` should print `1 2 3 4`.
- Inclusive ranges (`...`) aren't working. `echo {1...5}` should print `1 2 3 4 5`

drosseau:
FIxes #445

### DIFF
--- a/examples/braces.ion
+++ b/examples/braces.ion
@@ -1,1 +1,15 @@
 echo 1{A{1,2},B{1,2}}
+# inclusive ranges
+echo {-1...1}
+echo {2...0}
+echo {a...c}
+echo {d...b}
+echo {A...C}
+echo {D...B}
+# exclusive ranges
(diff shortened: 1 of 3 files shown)


mmstick:
Can you also add the corresponding tests in [here](https://github.com/redox-os/ion/blob/master/examples/braces.ion) and the output of that test [here](https://github.com/redox-os/ion/blob/master/examples/braces.out)? May want to add more tests in there too. Noticing that it's rather empty. You can run the test by running `sh examples/run_examples.sh`.

drosseau:
Yep sure I will add them in a bit.

mmstick:
Thanks!

PR #449, opened 2017-07-22
`expand_process` changed to not strip newlines for quoted instances

### LINKED ISSUE #410: Stripping newlines from subshell expansion
# Current Behavior
Grabbing the output of a command from an unquoted subshell expansion will strip the newlines from the output. For example:
```
$ ls -l
total 0
-rw-r--r-- 1 hunter hunter 0 Jul 13 12:26 bar
-rw-r--r-- 1 hunter hunter 0 Jul 13 12:26 baz
-rw-r--r-- 1 hunter hunter 0 Jul 13 12:26 foo
$ let var = $(ls -l)
$ let
...
var = total 0 -rw-r--r-- 1 hunter hunter 0 Jul 13 12:26 bar -rw-r--r-- 1 hunter hunter 0 Jul 13 12:26 baz -rw-r--r-- 1 hunter hunter 0 Jul 13 12:26 foo
...
```
# Proposal
I believe it is clearer if sub-shell expansion does not modify the output of the shell (some discussion in #405). Other thoughts?

huntergoldstein:
**Fixes**: Closes #410 

**Drawbacks**: If a user wants the specific behavior of capturing a process and stripping newlines, it is slightly more cumbersome to do so (as `tr "\n" " "` doesn't work as expected).

### DIFF
--- a/src/parser/mod.rs
+++ b/src/parser/mod.rs
@@ -46,7 +46,7 @@ macro_rules! get_expanders {
                         .map(|x| x.ascii_replace('\n', ' ').into())
                 }
             },
-            command: &|command: &str, quoted: bool| $vars.command_expansion(command, quoted),
+            command: &|command: &str| $vars.command_expansion(command),
         }
     }
 }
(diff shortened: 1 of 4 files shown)


mmstick:
Thanks!

PR #451, opened 2017-07-22
Simplify cross platform functions

jackpot51:
This set of changes moves most platform support into two simple files, `sys/redox.rs` and `sys/unix.rs`.

It also adjusts signal handling to use `signal` for compatibility with both Redox and Linux.

For the reviewer, probably @mmstick, I would like to know if signal behavior on Linux is the same as before this modification.


mmstick:
Thanks! From what I tested, it seems to be the same. Also reminds me that detection of the background operator is broken and needs to be fixed.

PR #460, opened 2017-07-23
Simplify signal logic so that it is safe and single-threaded

jackpot51:
This creates an atomic bitmask of pending signals, and removes the extra signal handling thread. When a signal arrives, an atomic operation flips a bit in the bitmask. The handling code unflips the bit when it checks for signals, in the main thread.


mmstick:
Thanks!

PR #464, opened 2017-07-25
Move OS #[cfg] functions to sys

### LINKED ISSUE #463: Move `shell::signals` and `shell::pipe::crossplat` to `sys`
Keeping all of the platform specific stuff under the `sys` module would be neater than having it spread around the project.

bb010g:
Resolves #463. A non-flat replica of the module structure may be wanted in the future if `sys` grows, but I think this works fine for now.


mmstick:
Thanks!

Issue #246, opened 2017-03-14
Associated Arrays (HashMaps)


mmstick:
Should support associative arrays to compete against Bash.

mmstick:
Here's a syntax we could use for associative array creation:

```ion
$ let hashmap = [ '&red=F00' '&blue=00F' ]
$ echo @hashmap['red']
FOO
```

And each associative array may contain further sub-elements:

```ion
$ let hashmap = [ '&user1={ &name="Michael" &id="1624" }' ]
$ echo @hashmap['user1']
> &name="Michael" &id="1624"
$ echo @hashmap['user1']['name']
> Michael
```

Sub-elements could be interpreted as arrays

```ion
$ let hashmap = [ '&users={&user1="1" &user2="2" &user3="3"}' ]
$ for user in @hashmap['users']
$   echo $user
$ end
> &user1="1"
> &user2="2"
> &user3="3"
```

Sub-elements may even be arrays themselves

```ion
$ let hashmap = [ '&team1=["12" "23" "21" "8"]' '&team2=["24" "8" "12" "30"]' ]
$ for score in @hashmap['team1']; echo $score; end
> 12
> 23
> 21
> 8
$ for score in @hashmap['team2']; echo $score; end
> 24
> 8
> 12
> 30
```

haraldh:
I can also imaging initializing a hashmap from a ```j#'<json>'#```:
```
let hashmap = j#' { "team1" : [ 12, 23, 21, 9 ], "team2" : [ 24, 8, 12, 30 ] }'#
let hashmap = j#' { 
   "team1" : [ 12, 23, 21, 9 ], 
   "team2" : [ 24, 8, 12, 30 ] 
}'#
let hashmap = j#' { "users" : { "user1" : 1, "user2" : 2,  "user3" : 3} }'#
let hashmap = j#' {
   "users" : { 
      "user1" : 1, 
      "user2" : 2,  
      "user3" : 3
   }
}'#
```
this way, you could reuse serde::json and the serde_json::Value enum.

or a yaml string ```y#'<yaml>'#```
```
let hashmap = y#'
team1:
- 12
- 23
- 21
- 9
team2:
- 24
- 8
- 12
- 30
'#
let hashmap = y#'
users:
  user1: 1
  user2: 2
  user3: 3
'#
```

mmstick:
I'm hesitant to use YAML. JSON is more likely, but RON is also of interest.

PR #465, opened 2017-07-25
Implement simple form of associative arrays

bblancha:
**Problem**: No map-like data structures yet exist in ion.

**Solution**: Allow simple associative arrays with string keys

- Create hashmap context in shell::variables to store named hashmaps
- Create "Key" selector to represent string keys
- Instances of "Select" objects must now be cloned; remove Copy trait
      (This was needed to support a select type based on string keys)

**Drawbacks**: Select objects no longer implement Copy

**TODOs**: Does not yet have an initializer syntax for associative arrays 

**Fixes**: Addresses #246 

**State**: Needs review


mmstick:
Thanks!

PR #466, opened 2017-07-26
Compile on Darwin

llambda:
Make compilable on Mac. I don't think it has a pipe2 system call.


mmstick:
This will work for now, but there will be an issue with background job control not working properly, because of the lack of the `WCONTINUED` flag. The platform supports it, it's just that the `nix` crate isn't allowing you to use the flag because it's not also defined in `nix` for that platform.

llambda:
Ah, interesting.

PR #468, opened 2017-07-26
include whitespace in alias argument

### LINKED ISSUE #453: Alias Argument Concatenation Issue
Currently, `alias ls = ls --color` will create a `ls` alias that contains `ls--color`. Just a simple problem of missing whitespace between arguments.

nivkner:
Fix #453 
The `alias` built-in took an array of arguments so whitespace wasn't included. 
Now the array is joined by whitespace, and given to `alias` as is. 
`parse_assignment` didn't actually parse `let` bindings so I changed the name and documentation to reflect that.


mmstick:
Thanks!

Issue #364, opened 2017-07-05
Windows Support Status


mmstick:
# Problems with Supporting Windows

Take the `fork` syscall, for example. The shell needs to be able to perform forking in order to create background processes that are an exact copy of the parent process but executing a different codepath, and to assign new process group IDs to these background child processes. Windows does not support forking as a concept at all, so we'd need to replicate the forking concept from scratch for Windows. It's one of the biggest reasons why most system shells are *nix-exclusive.

There's also the manner in which *nix architectures perform job control, of which Windows also does not support at all. With *nix architectures, you configure spawned processes to set their process group ID to be the same as their process ID, and manage processes with various *nix signals, and use functions like `tcsetpgrp` to set foreground processes.

Or the difference between how *nix systems operate on file descriptors, which are integers, but Windows operates on handles, which are void pointers. Jobs are controlled by passing along PIDs and mapping FDs on *nix systems, but Windows is very different.

# What Needs To Be Done

Basically, to gain Windows support, all of the code that manages foreground and background processes, process creation and execution, and piping / redirection would need to be rewritten from scratch with Windows-specific APIs. But for that to happen, we first need someone who uses and develops for Windows to be willing to do that.

# Checklist

- [ ] Piping
- [ ] Redirection
- [ ] Signal Handling
- [ ] Replicate Forking Functionality
- [ ] Handle-based Job Management
- [ ] Process Groups
- [ ] Background & Foreground Process Management
- [ ] Integrate AppVeyor for Testing Windows Builds

huntergoldstein:
Not only do we need to implement the above, but do so in a readable and maintainable way.

mmstick:
I'd just ship Windows-specific routines in their own modules. A start would likely be just taking the `pipe.rs` module, making that a `pipe` directory, and then having a `unix.rs` and `windows.rs` file within. Maybe even `redox.rs` for the Redox-specific bits, although Redox largely shares a lot of unix functionality.

huntergoldstein:
I think the bits that Redox and Unix would share are the same as Unix and Windows. What's not shared is the specifics for setting a pipe, forking, and the different signal types.

PR #469, opened 2017-07-26
Allow builtins to be executed as part of pipelines

### LINKED ISSUE #379: Execute Builtins Within Pipeline Execution
If a builtin needs to be piped, we should do precisely that. We should achieve this by creating new file descriptors, and mapping the stdout and stderr of these builtin commands to these file descriptors.

### LINKED ISSUE #454: Parsing Issue w/ && Operator
I'm not sure why this is at the moment, but:

```sh
matches Foo '([A-Z])\w+' && echo true
```

Gives the following output:

```
ion: syntax error: '(' at position 13 is out of place
```

But this:

```sh
matches Foo '([A-Z])\w+'; echo true
```

Gives:

```
true
```

I've also noticed that this:

```sh
matches Foo '[A-Z]\w+' && echo true
```

Gives no output, but this does:

```sh
matches Foo '[A-Z]\w+'; echo true
```

Likely the same issue, and likely in the `StatementSplitter` logic.

huntergoldstein:
**Problem**: Currently if we have a builtin to be run as part of a pipeline, we just take the job, reconstruct it, and run a sub-process such as `ion -c builtin arg1 arg2`. This is problematic as it forces `ion` to re-parse the given arguments which is inefficient and bug prone ( #454 ).

**Solution**: Manually manage running builtins as part of a pipeline by forking the current process and executing the builtin in the child process.

- Replaced the use of `Command` in `shell::pipe` with `RefinedJob`; an enumeration that is either an external command or a builtin
- Modified `shell::pipe::pipe` to fork and execute a builtin for the `RefinedJob::Builtin` instance; behavior for `RefinedJob::External` is roughly the same as before: build a `Command`, run `spawn()`.

**Drawbacks**: The use of `RawFd` everywhere might make #364 more difficult.

**Fixes**: Closes #379 ; Closes #454 

**State**: WIP; I want to make sure there are some regression tests in `examples/`. Also this is a 500+- change so letting it sit for review would be good.

**TODO**:
- [x] Determine the redox equivalent for `STDOUT_FILENO` et al
- [x] Add a regression example in `examples/` for piping to / from a builtin #

### DIFF
--- a/examples/builtin_piping.ion
+++ b/examples/builtin_piping.ion
@@ -1,0 +1,6 @@
+matches Foo '([A-Z])\w+' && echo true
+matches foo '([A-Z])\w+' || echo false
+
+fn foobar x; end
+
+fn | tr '[a-z]' '[A-Z]'
--- a/examples/builtin_piping.out
+++ b/examples/builtin_piping.out
@@ -1,0 +1,4 @@
+true
+false
+# FUNCTIONS
+    FOOBAR
--- a/src/shell/fork.rs
+++ b/src/shell/fork.rs
@@ -51,7 +51,7 @@ mod unix {
 }
 
 use std::process::{Command, exit};
-use super::job::JobKind;
+use super::job::{RefinedJob, JobKind};
 use super::job_control::{JobControl, ProcessState};
 use super::Shell;
 use super::signals;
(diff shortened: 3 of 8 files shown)


huntergoldstein:
Also it seems I forgot to add `STDOUT_FILENO` for redox, so definitely WIP.

mmstick:
Looks good to me so far.

huntergoldstein:
Good to go; I wanted to add an example for `read`, but it doesn't work when `stdin` is not a tty.

mmstick:
Thanks for this work!

PR #481, opened 2017-07-27
Fixed job arguments expanding to empty string

### LINKED ISSUE #345: empty string arguments

huntergoldstein:
- If a job argument expands to the empty string, return an array containing the empty string instead.
- Added a regression test w.r.t. job expansion

**Fixes**: Closes #345 (again).

------

I wonder if it would be helpful to make a distinction inside Ion's code-base between string and array values. I.e. `Array` becomes `Value`, and `Value` becomes an enum like:
```
enum Value {
  String(SmallString<...>),
  Array(SmallVec<...>),
}
```
Though we probably can't use `SmallVec`; it would have to be `Vec` (I think).

### DIFF
--- a/src/shell/job.rs
+++ b/src/shell/job.rs
@@ -30,7 +30,12 @@ impl Job {
         let mut expanded = Array::new();
         expanded.grow(self.args.len());
         expanded.extend(self.args.drain().flat_map(|arg| {
-            expand_string(&arg, expanders, false)
+            let res = expand_string(&arg, expanders, false);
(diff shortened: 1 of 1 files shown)


mmstick:
It would be better if we could make that distinction. Thanks for fixing this though!

PR #484, opened 2017-07-28
Switch from HISTORY_FILE to HISTFILE

### LINKED ISSUE #299: use HISTFILE instead of HISTORY_FILE
There's probably some context I'm missing here, but shouldn't the shell be using "HISTFILE" instead of "HISTORY_FILE"? HISTFILE is the standard bash variable:

https://www.gnu.org/software/bash/manual/html_node/Bash-Variables.html

bb010g:
Resolves #299


mmstick:
Thanks!

PR #485, opened 2017-07-28
read builtin detects if `stdin` is a TTY

### LINKED ISSUE #474: Allow `read` builtin to take input from non-tty `stdin`
**Reproduction**: Attempt to have `read` take input from anything except a terminal, for example:
```
$ echo foo | read bar
```
**Expected behavior**:
```
$ echo foo | read bar
$ echo $bar
foo
```

**Actual behavior**: 
```
$ echo foo | read bar
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Custom(Custom { kind: Other, error: StringError("Unable to get Termios attribute.") }) }', /checkout/src/libcore/result.rs:860:4
$ echo $bar
<no output>
```

**Build information**: 
- `git rev-parse HEAD`: 2c62ffb8ff8c20aa442c6c3ae9d7f0397baf1460

------

The solution to this is to have the `read` builtin determine if its `stdin` is a tty or not; if so, it will use the nice `liner` prompt, otherwise it will read the next line from `stdin`.

huntergoldstein:
- Added `sys::isatty` which emulates `libc`'s `isatty`
- `read` builtin detects if `stdin` is a TTY: if so, it uses `liner`, otherwise it reads line by line assigning each new variable to each line

**Drawbacks**: If we have input like:
```
Foo
Bar
Baz
```
and call
```
$ read foo bar < input
```
The last line of input will be lost. Is this desirable? Or should we make like Bash's `read` which uses the last variable as the sink for remaining input. If so, we could make it an array of lines.

**Fixes**: Closes #474. Disclaimer: _this does not resolve all issues mentioned in the issue_ i.e.:
```
echo "foo" | read bar
```
will do nothing as `read bar` is executed in a subshell. 

**State**: Good to go barring the issue mentioned in **Drawbacks**

### DIFF
--- a/examples/builtin_piping.ion
+++ b/examples/builtin_piping.ion
@@ -4,3 +4,6 @@ matches foo '([A-Z])\w+' || echo false
 fn foobar x; end
 
 fn | tr '[a-z]' '[A-Z]'
+
+read foo <<< $(echo bar)
+echo $foo
--- a/examples/builtin_piping.out
+++ b/examples/builtin_piping.out
@@ -2,3 +2,4 @@ true
 false
 # FUNCTIONS
     FOOBAR
+bar
--- a/src/shell/variables.rs
+++ b/src/shell/variables.rs
@@ -1,5 +1,6 @@
 use fnv::FnvHashMap;
 use std::env;
+use std::io::{self, BufRead};
 use std::process;
 
 use super::directory_stack::DirectoryStack;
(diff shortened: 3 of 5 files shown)


mmstick:
Thanks! I think it'd be okay to use the last variable as an array that contains remaining lines. Maybe the exit status could be used to determine if the last variable was initialized as an array or string?

huntergoldstein:
The exit status ought to be reserved for "did this program run successfully." However we are planning on an `exists` builtin that determines if a given variable exists as a string / array no?

mmstick:
We are planning on that, indeed.

Issue #424, opened 2017-07-16
Subshells Should Fork


mmstick:
Currently, Ion pre-expands subshell expressions before creating a new Ion process and handing it the expanded form. Instead, we should properly fork the shell at this point, and get rid of all the logic for handling shell expansion pre-expansions. This means that, currently, the following doesn't work:

```ion
fn test; echo testing; end
echo $(test)
```

drosseau:
I'm trying to find a nice fix for this but it is getting kinda messy. I think the problem is complicated by the fact that expansion isn't handled by the `Shell` but by a bunch of closures over the variables and directory stack, so when you get to `command` you have no access to the `Shell`. I think it'd make this easier and make more sense if the expand functions functions were actually implemented by `Shell` so we didn't have to use the closures and the `get_expanders!` macro. That'd be a lot of refactoring though.

huntergoldstein:
I believe that's what `ExpandFunctions::command` is for no? It might be nice to encapsulate the subshell logic so that running methods / builtins in forks can share all that same code.

I played around with making `ExpandFunctions` a trait instead of a struct. To me it _feels_ better* and it wouldn't require a ton of refactoring.

* Read: more idiomatic in my admittedly limited experience

drosseau:
`command` doesn't know about any of the user defined functions and just creates a new ion process to run the command, similar to how builtins were working before. That is why his example isn't working. I played around with the same thing but, I ended up needing to do a decent amount of refactoring and scrapped my approach. Haven't had time to look at it for a while though.

mmstick:
Access can be granted to builtins and functions by just adding their associated fields to the expander.

huntergoldstein:
Making `ExpanderFunctions` into an `Expander` trait might be a slightly nicer way such that `command` can know about builtins / functions (by passing, say, `Shell` as `&self`). 

An intermediate step could be to have `ExpanderFunctions` implement `Expander` as many tests create their own special instances of `ExpanderFunctions`. I tried this, but it felt messy and I'd rather just scrap `ExpanderFunctions` (and especially `get_expanders!` altogether).

huntergoldstein:
I've been hacking away at the above for a good bit of today; getting rid of `ExpanderFunctions` altogether is a little tougher as a lot of tests use their own special versions, but I think a sensible thing to do is to have any `Expander` trait have its functions return `None` by default then each test can implement something like:
```
struct LocalExpand;

impl Expander for LocalExpand {
[...]

PR #488, opened 2017-08-02
Replace `ExpanderFunctions` with `Expander` trait

huntergoldstein:
**Problem**: `ExpanderFunctions` was somewhat unruly as:
- It was mostly used through `get_expanders!` which made tracking down errors more cumbersome as it was generated through a macro
- Hard to extend sensibly as the main instance, `get_expanders!`, would lose access to any information that was not contained within `Variables` or `DirectoryStack`
- Had weird artifacts like having a direct reference to a `Variables` but also a separate `variable(&str, bool) -> Option<Value>` function?

**Solution**: Replace `ExpanderFunctions` with an `Expander` trait which provides roughly the same interface as `ExpanderFunctions` but allows the backing implementation to be more complex 

- Removed `ExpanderFunctions`, `get_expanders!` etc.
- Add the aforementioned `Expander` trait
- Have `Shell` implement `Expander` and use `Shell` as the `Expander` for most non-testing instances
- Refactored `let_assignment` and `export_assigment` into `VariableStore` trait, which represents storing a `Binding` as either a local variable or a global (exported) variable
  - This was done as `let_assignment` and `export_assignment` took in a mutable `Variables`, but also required an `ExpanderFunctions` to expand the right hand side. Instead, `Shell` now implements `VariableStore`

**Drawbacks**: I am not sure this is as performant as `ExpanderFunctions` but its more idiomatic IMO.

**Fixes**: This will make #424 much much easier.

**State**: Good to go pending review!

### DIFF
--- a/src/builtins/variables.rs
+++ b/src/builtins/variables.rs
@@ -210,12 +210,15 @@ pub fn drop_variable<I: IntoIterator>(vars: &mut Variables, args: I) -> i32
 #[cfg(test)]
 mod test {
     use super::*;
-    use parser::{expand_string, ExpanderFunctions, Select};
+    use parser::{expand_string, Expander};
     use shell::status::{FAILURE, SUCCESS};
-    use shell::directory_stack::DirectoryStack;
 
-    fn new_dir_stack() -> DirectoryStack {
-        DirectoryStack::new()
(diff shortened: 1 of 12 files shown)


mmstick:
Thanks!

PR #499, opened 2017-08-15
Reformatted Manual based on Rust book (2ed.)

jwarner112:
**Problem**: The manual has structural issues that I'll continue to dive into, but the most pressing was having broken links on the Introduction page (pointing to `.md` files instead of built HTML files).

**Solution**: Restructured in a flat way (based on Rust book 2nd Edition) that `mdbook` parses corectly.

- All nested content e.g. pages under `flows/` were moved up a level.
- All pages with bulleted lists but no pages for them were segregated to make pages for said list items.

**Drawbacks**: Organization of the manual is of personal taste and this is just my first pass over it; I'll revise it some more later for content and clarity.


mmstick:
Thanks!

PR #501, opened 2017-08-16
Fix syntax error in src/shell/variables.rs

BafDyce:
introduced by c1fdd5e

### DIFF
--- a/src/shell/variables.rs
+++ b/src/shell/variables.rs
@@ -200,7 +200,7 @@ impl Variables {
                         eprintln!("ion: {} is not a valid color", name);
                         None
                     }
-                }
+                },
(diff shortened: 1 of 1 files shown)


BafDyce:
Note: I just discovered that this "syntax error" is not a syntax error in rust-nightly (I had `rustc 1.19.0 (0ade33941 2017-07-17)` when I encountered the compilation error, with `rustc 1.21.0-nightly (df511d554 2017-08-14)` it compiles fine.

As ions readme does not state that one should use the nightly compiler (this is only stated in redox' readme), I don't know whether ion is intended to be compiled with nightly or without.

mmstick:
Probably should update our Travis file to build with the stable compiler.

Issue #409, opened 2017-07-13
List of Builtins to Implement


mmstick:
# Conditional Builtins

- `not <command>`: Reverses the exit status value of the given command.
- `contains $key @values`: Returns 0 exit status if the key is contained within all values
- `intersects $key @values`: Similar to the contains builtin -- but only requires one match
- `matches $key $regex`: A regex-based evaluator
- `isatty [FD]`: return true exit status if the supplied FD is a tty 

# Math

- `random`: generate random numbers

# I/O
- `printf`: Implement the same as other shells
- `printr`: Rusty variant maybe?
- `cat`: Perhaps copy `cat` from coreutils into Ion.

suhr:
> md5sum

I'd rather avoid having this as a builtin. It is neither secure nor really fast.

mmstick:
@suhr md5 checksums aren't for security purposes though.

suhr:
Anyway, there's a lot of various hashes, and it's better to avoid making one of them more special than others, especially when that one is not the best in the class.

bb010g:
`exec`?

mmstick:
@suhr: What would you recommend, if any?
@bb010g: Adding that to the list

bb010g:
Looks like we're also missing a `source` builtin, unless you were planning on doing something special with external functions. (If it's feasible, being able to source into a namespace would be amazing.)

mmstick:
@bb010g [The source builtin is already implemented](https://github.com/redox-os/ion/blob/master/src/builtins/source.rs). It was one of the first builtins to be implemented when development started.

bb010g:
Sorry, somehow missed it in the documentation.

drosseau:
Aren't most of the `exists` functions just duplicates of different `test` flags?

`exists -f FILE` -> `test -f FILE`
`exists -d PATH` -> `test -d PATH`
`exists -s STRING` -> `test -z STRING`

I don't think test covers the first one though. Is that kinda like `command -v`?

mmstick:
There is some overlap there, yes; and it's purpose is the typical purpose of using `command -v`.  I've thought about expanding our variant of `test` to also include binary checks, but I also don't really like the `test` name.

mmstick:
Other possibly useful flags are:

```
exists -v var_name
exists -a array_name
```

huntergoldstein:
[...]

PR #504, opened 2017-08-16
Add "exists" builtin

BafDyce:
**Problem**: Implement the `exists` builtin mentioned in #409 

**Solution**:
I copied over the `test` builtin implementation, removed the stuff we don't need in `exists` and implemented the necessary new logic.

- Added `exists` builtin
- Implemented the following flags: `-a`, `-b`, `-d`, `-f`, `--fn`, and `-s`
- Implemented the STRING check
- Implemented unit tests for all functions
- Implemented integration tests

**TODOs**:
- Maybe: `binary_is_in_path` should use the same logic as the one that is used to spawn new processes
- `test_binary_is_in_path` could benefit from some more unit tests, however this requires a more complex/enhanced `testing/`-directory structure

**Fixes**: #409 (partly)

**State**: ready

**Blocking/related**: #409

**Other**:

------

_The above template is not necessary for smaller PRs._

### DIFF
--- a/examples/exists.ion
+++ b/examples/exists.ion
@@ -69,6 +69,11 @@ echo $?
 exists -f does-not-exist
 echo $?
 
+fn testFunc a
+  echo $a
+end
+exists --fn testFunc
+echo $?
 
 exists -s
 echo $?
--- a/examples/exists.ion
+++ b/examples/exists.ion
@@ -1,0 +1,91 @@
+# Same tests as in src/builtins/exists.rs:test_evaluate_arguments()
+# TODO: find a better way than to write "echo $?" between each test cases
+exists
+echo $?
+exists foo bar
+echo $?
+
+exists --help
+echo $?
+exists --help unused params
+echo $?
+
+
+exists ""
+echo $?
+exists string
+echo $?
+exists "string with spaces"
+echo $?
+exists "-startswithdash"
+echo $?
+
+
+exists -a
+echo $?
+let emptyarray = []
+echo @emptyarray
+exists -a emptyarray
+echo $?
+let array = [ "element" ]
+echo @array
+exists -a array
+echo $?
+exists -a array
+echo $?
+
+
+exists -b
+echo $?
+let OLDPATH = $PATH
+let PATH = testing/
+echo "PATH = $PATH"
+ls -1 $PATH
+exists -b executable_file
+echo $?
+exists -b empty_file
+echo $?
+exists -b file_does_not_exist
+echo $?
+let PATH = $OLDPATH
+echo "Reset PATH to old path"
+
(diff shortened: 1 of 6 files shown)


mmstick:
It's looking perfectly fine to me so far. Tell me when you're finished and I'll merge it.

mmstick:
Also, I would require that flags have spaces between the argument that follows, so there won't be any confusing about strings that start with dashes.

BafDyce:
Currently, the implementation returns true/success when it does not recognize the flag or if a flag is provided without an argument. So, if only a single argument is given which starts with a dash, `exists` always returns success. Is this ok? Or should it return false if a valid flag is given without an argument?

Besides this, I would say the the PR is finished. The Todos which are still open are not straight forward to implement and don't negatively affect the functionality right now:
* unit tests for `evaluate_arguments`: I don't really know what/how I should test here.
* binary_is_in_path should use the same logic as the one that is used to spawn new processes; However the logic for spawning new process relies on spawning them and returns an error if this failed. This approach is not suitable for `exists`
* more tests for `test_binary_is_in_path`, however this requires a more complex directory structure of `testing/` (sub-directories with different permissions, etc.)

BafDyce:
Also, should I rebase the branch onto the latest master or is it ok to keep it as it is as long as there is no merge conflict?

BafDyce:
**unit tests:**
 I know that (I used it for some of the other test functions). I just don't know what I should test for. Especially, I don't know how i should check whether an invocation with --help printed the help page.

**behavior**

Ok, then I will do this.

mmstick:
> Especially, I don't know how i should check whether an invocation with --help printed the help page.

This sort of functionality cannot be tested in unit tests, but you can test it with integration tests, which are located in the examples directory.

BafDyce:
So. Right now, `exists` still returns success for invalid flags and for flags without arguments. In this cases we just treat the single provided argument as a string.

If we wanted to return failure for invalid flags and for valid flags without another argument, then we would need another way to handle strings that start with a dash. A way for this would be to add a `--` flag, like this:

```rust
// fn match_flag_argument()
match flag {
        'a' => array_var_is_not_empty(argument, shell),
        'b' => binary_is_in_path(argument, shell),
        'd' => path_is_directory(argument),
        'f' => path_is_file(argument),
        's' => string_var_is_not_empty(argument, shell),
        '-' => string_is_nonzero(argument),
        _ => false,
    }
```

This would still leave an open issue: How to detect & handle invalid flags that have more than one letter, without adding too much biolerplate?

BafDyce:
From my pov, the PR is ready to merge.

mmstick:
Thanks! I will go ahead and merge it then.

Issue #490, opened 2017-08-05
Time Command Should Time Functions & Builtins


mmstick:
The current `time` builtin is very simplistic, and only supports timing external commands.

chrisvittal:
I've been looking into this. 

I'm wondering if there is any way to put the `args` back into the the shell and execute them there, or if that is even the best solution. Do we already have the full parse tree for the `args` here?

I don't know how `time` is implemented in other shells, however I feel like we shouldn't be timing how long it takes ion to join and re-parse the arguments time has been given.

Thanks!

mmstick:
I believe the solution would be to make `time` a keyword, rather than a builtin command.

Statements are parsed & mapped in `src/parser/statement/parser.rs`. We'd just need to add `Time` as a Statement keyword on it's own. The logic for handling these variants is in `src/shell/flow.rs` -- we'd just treat the `Time` variant as if it were a regular pipeline.

Something to consider though is that because the time command can time other keywords, the `Time` variant should contain a `Statement`.

PR #505, opened 2017-08-17
Change time from builtin to shell keyword.

chrisvittal:
**Problem**: The `time` command could not accept shell commands like builtins or for loops. #490 

**Solution**: Refactored `time` from a builtin to a keyword. That is, create a `Statement::Time` and
handle the `time` logic in the execution of `Statement::Time` variants, rather than as a builtin function.

 - A `Time` variant of `Statement` a boxed statement member.
 - Support for parsing the time keyword in `parser/statement/parse.rs` by
   recursing on the rest of the input to the `parse` function when time is
   encountered.
 - A port of the logic from the old builtin into the `execute_toplevel`
   impl for `Shell`, where instead of using the error handling from
   Command, the time statement is executed through a recursive call to
   `execute_toplevel`.
- The removal of all references to the old `time` builtin.

**Drawbacks**: The loss of some friendliness. The `time -h` invocation no longer works.

**TODOs**: 
- Add tests.
- Time currently cannot handle multiline statements like quotes being spread over multiple lines. Fix this.
- Remove `builtins/time.rs` file
- Edit: 2017-08-23 Implemented at current, a `time` statement inside of a loop supresses all output. Fix this.

**Fixes**: #490 

**State**: WIP


mmstick:
I think I'll make a new release of Ion's alpha as soon as this and #504 is merged. Or if any other PR's come up during this timeframe, I'll wait until they are merged too.

mmstick:
Are you still working on this?

chrisvittal:
I am.

Haven't had much time since I've opened the PR. Working on getting it to wait until a statement is actually being executed to time things, rather than timing nothing if the run level needs to be increased (like with a `for` loop).

I did a little investigating other `time` commands in the meantime and here's what I've found for other shells timing things that could take multiple statements:

- `bash`: Works well, waits until there is something to execute and then times the execution.
- `zsh`: Fails silently. No timing is displayed on a multiline `time for` type statement.
- `time` program: Doesn't handle shell builtin or other control constructs at all, but I wouldn't expect it to.

I need a bit of assistance in getting what I have to work with multiline statements. I'm also not sure what should be done for things like the following:

```
for var in @(filter_cmd)
    time long_running_command $var
end
```
It seems like I may need to add more stuff to `Shell`, like `in_time_command` and `current_time_runlevel`, to create some kind of bookeeping around all of this.

Thanks!

chrisvittal:
This is feature complete. If you were waiting on me, I apologize. Please take a look.

Thank you!

mmstick:
Thanks!

Issue #503, opened 2017-08-16
Add environment variable to define commands that should NOT be saved in the history


BafDyce:
For example, bash (and probably other shells too) ignores commands starting with a space when `HISTCONTROL` is set to `ignorespace` or `ignoreboth` (see www.delorie.com/gnu/docs/bash/bashref_62.html). This would be a useful feature and is part of my current daily workflow.

For ion, we could add a variable such as `HISTORY_IGNORE` which contains rules to influence which commands are saved to the history.

**Blocking/related**: #502 

**Misc**:
I'd be happy to implement this feature, however I'd like to have it discussed it first:
* Do you/the community think this is a useful addition?
* Is `HISTORY_IGNORE` a good name?
* Should the "ignore-character" always be a space? Any whitespace? Any user-defined character?
* What's a sane way to combine this with the issue mentioned in #502?

**Note**:
I've created two separate issues, feel free to merge them if you think that would be better.

------

_If the above does not fit the nature of the issue feel free to modify it._

mmstick:
- I think the name would be fine, or we could even just use the same name as Bash.
- For simplicity, we should probably simply have it as any whitespace.
- As for #502, it would be ideal if we could just have a temporary history buffer that stores the last issued command in memory, whether it failed or not, without committing it to the disk for long term calling. However, you'd need to delve into the liner crate to figure out how to implement that.

BafDyce:
I started implementing this feature today. However, I am not quite sure what the ` || self.control_flow.level > 0` check in [src/shell/binary.rs, line 278](https://github.com/redox-os/ion/blob/master/src/shell/binary.rs#L278) is for.

This condition will only be checked when `self.previous_status` equals `NO_SUCH_COMMAND`. And if I understand the logic in [src/shell/flow_control.rs](https://github.com/redox-os/ion/blob/master/src/shell/flow_control.rs) correctly, `self.control_flow.level` is only greater than zero if we're inside a function, if/else block, match body, etc. Therefore, I don't get this additional check at all. Is someone able to explain this to me?

**@ name** I'll use `HISTORY_IGNORE` as this is consistent with  `HISTORY_SIZE`. Also, I'd like to use a different name than in bash as the syntax/keywords will be different. However, we could also make it consistent with `HISTFILE*` env variables and name the variables `HISTSIZE` and `HISTIGNORE`.

**@ whitespace**: sounds like a reasonable default. I will use this.

[...]

Issue #502, opened 2017-08-16
Only commands with exit status != `NO_SUCH_COMMAND` are saved in history?


BafDyce:
**Reproduction**: enter a command which is not found, e.g.  `gerp -rn "some_complex_regex" .`; note the typo in `grep`

**Expected behavior**: Even though the command was not found, it's invocation attempt should be stored in the history, so that one can press `arrow up`, fix the mistake, and rerun the command.

```
# echo hi
hi
# gerp -rn "some_complex_regex" .
ion: command not found: gerp
# history | tail -1
gerp -rn "some_complex_regex" .
```

**Actual behavior**:
```
# echo hi
hi
# gerp -rn "some_complex_regex" .
ion: command not found: gerp
# history | tail -1
echo hi
```
**Build information**:
```
# rustc -V
rustc 1.19.0 (0ade33941 2017-07-17)
# git rev-parse HEAD
5bbc65d55f1eb75e89eb5141d51b57d0eecd91c8
# uname -a
Linux <hostname> 4.12.6-1-ARCH #1 SMP PREEMPT Sat Aug 12 09:16:22 CEST 2017 x86_64 GNU/Linux
```

**Misc**:
I understand that the current behavior might be desired for some people and it probably is a sane default. However, if the command was rather complex/long its a pain to a) retype or b) copy & paste it with the mouse.
Maybe, it would be possible to add an environment variable to modify this behavior? What would be a good name for that variable? `HISTORY_IGNORE` maybe?

------

_If the above does not fit the nature of the issue feel free to modify it._

bb010g:
Was this fixed by #512?

mmstick:
That's correct. This was fixed.

PR #512, opened 2017-08-23
WIP: Allow user to prevent commands from being saved in history

BafDyce:
Commands that should NOT be saved in the history can be defined via the
`HISTORY_IGNORE` environment variable (array). Currently, the following
options are supported:
- `"all"`: No commands will be saved until this option is removed again
- `"whitespace"`: ignores commands with a leading whitespace character
- `"no_such_command"`: ignores commands with exit status NO_SUCH_COMMAND
- `"regex:xxx"`: where xxx is a regex (https://doc.rust-lang.org/regex), ignores all commands that match this regex

**Example**:
```
# echo 5
5
# echo 4
4
# echo 3
3
# echo 2
2
# echo 1
1
# history | tail -5
echo 5
echo 4
echo 3
echo 2
echo 1
# let HISTORY_IGNORE = [ "all" ]
# true
# false
# history | tail -5
echo 3
echo 2
echo 1
history | tail -5
let HISTORY_IGNORE = [ "all" ]
# let HISTORY_IGNORE = [ "whitespace" ]
# history | tail -5
echo 2
echo 1
history | tail -5
let HISTORY_IGNORE = [ "all" ]
let HISTORY_IGNORE = [ "whitespace" ]
# true
#  false
# history | tail -5
history | tail -5
let HISTORY_IGNORE = [ "all" ]
let HISTORY_IGNORE = [ "whitespace" ]
history | tail -5
true
# let HISTORY_IGNORE = [ "no_such_command" ]
# history | tail -5
let HISTORY_IGNORE = [ "whitespace" ]
history | tail -5
true
history | tail -5
let HISTORY_IGNORE = [ "no_such_command" ]
# true
#  false
# trulse
ion: command not found: trulse
# history | tail -5
history | tail -5
let HISTORY_IGNORE = [ "no_such_command" ]
history | tail -5
true
 false
# let HISTORY_IGNORE = [ "whitespace" "regex:false" ]
# true
#  true
# false
#  false
# history | tail -5
true
 false
history | tail -5
let HISTORY_IGNORE = [ "whitespace" "regex:false" ]
true
# let HISTORY_IGNORE = [ "whitespace" "regex:\d" ]
# true
# true 1
# false
# false 2
# history | tail -5
true
history | tail -5
let HISTORY_IGNORE = [ "whitespace" "regex:\d" ]
true
false
# let HISTORY_IGNORE = [ "whitespace" "regex:a{2}" ]
# true
# true a
# true aa
# true aaa
# history | tail -5
false
history | tail -5
let HISTORY_IGNORE = [ "whitespace" "regex:a{2}" ]
true
true a
# let HISTORY_IGNORE = [ "whitespace" regex:[1-2] ]
# true a
# true 122
# true 2
# true 3
# history | tail -5
true a
history | tail -5
let HISTORY_IGNORE = [ "whitespace" regex:[1-2] ]
true
true 3
# let HISTORY_IGNORE = [ "whitespace" regex:[[:alpha:]][[:space:]][[:digit:]] ]
# echo abc123
abc123
# echo abc 123
abc 123
# echo 123 abc
123 abc
# echo def  456 # 2 spaces in between
def  456
# history | tail -5
history | tail -5
let HISTORY_IGNORE = [ "whitespace" regex:[[:alpha:]][[:space:]][[:digit:]] ]
echo abc123
echo 123 abc
echo def  456
# let HISTORY_IGNORE = [ "regex:#ignore" ]
# true
# false #ignore
# history | tail -3
history | tail -5
let HISTORY_IGNORE = [ "regex:#ignore" ]
true
```

**Solution**:
I implemented it in a way that is open for future extensions, unlike `bash`' implementation ;)

**TODOs**:
- documentation
- unit & integration tests

**Fixes**:
#503
#502 (partly, see https://github.com/redox-os/ion/issues/503#issuecomment-323079748)

**State**:
**WIP** This might take me some days though, as I'll be quite busy the next days.

**Other**:
Also open to discussion if you think I could implement something differently, etc.

### DIFF
--- a/src/shell/history.rs
+++ b/src/shell/history.rs
@@ -19,6 +19,21 @@ bitflags! {
     }
 }
 
+pub struct IgnoreSetting {
+    flags: IgnoreFlags,
+    // Yes, a bad heap-based Vec, however unfortunately its not possible to store Regex'es in Array
+    regexes: Option<Vec<Regex>>,
+}
+
+impl IgnoreSetting {
+    pub fn default() -> IgnoreSetting {
+        IgnoreSetting {
+            flags: IgnoreFlags::empty(),
+            regexes: None,
+        }
+    }
+}
+
 /// Contains all history-related functionality for the `Shell`.
 pub trait ShellHistory {
     /// Prints the commands contained within the history buffers to standard output.
@@ -39,14 +54,14 @@ pub trait ShellHistory {
     /// Saves a command in the history, depending on @HISTORY_IGNORE. Should be called
     /// immediately after `on_command()`
     fn save_command_in_history(&mut self, command: &str);
+
+    /// Updates the history ignore patterns. Call this whenever HISTORY_IGNORE is changed.
+    fn update_ignore_patterns(&mut self, patterns: &Array);
 }
 
 trait ShellHistoryPrivate {
     /// Returns true if the given command with the given exit status should be saved in the history
     fn should_save_command(&self, command: &str) -> bool;
-
-    /// Parses the @HISTORY_IGNORE environment variable
-    fn parse_history_ignore(&self) -> (IgnoreFlags, Option<Array>);
 }
 
 impl<'a> ShellHistory for Shell<'a> {
@@ -99,11 +114,44 @@ impl<'a> ShellHistory for Shell<'a> {
             }
         }
     }
+
+    fn update_ignore_patterns(&mut self, patterns: &Array) {
+        let mut flags = IgnoreFlags::empty();
+        let mut regexes = Vec::new();
+        // for convenience and to avoid typos
+        let regex_prefix = "regex:";
+        for pattern in patterns {
+            match pattern.as_ref() {
+                "all" => flags |= IGNORE_ALL,
+                "no_such_command" => flags |= IGNORE_NO_SUCH_COMMAND,
+                "whitespace" => flags |= IGNORE_WHITESPACE,
+                // The length check is there to just ignore empty regex definitions
+                _ if pattern.starts_with(regex_prefix) && pattern.len() > regex_prefix.len() => {
+                    flags |= IGNORE_BASED_ON_REGEX;
+                    let regex_string = pattern[regex_prefix.len()..].to_owned();
+                    // We save the compiled regexes, as compiling them can be  an expensive task
+                    if let Ok(regex) = Regex::new(&regex_string) {
+                        regexes.push(regex);
+                    }
+                }
+                _ => continue,
+            }
+        }
+
+        self.ignore_setting.flags = flags;
+        self.ignore_setting.regexes = if regexes.len() > 0 {
+            Some(regexes)
+        } else {
+            None
+        }
+    }
 }
 
 impl<'a> ShellHistoryPrivate for Shell<'a> {
     fn should_save_command(&self, command: &str) -> bool {
-        let (ignore, regexes) = self.parse_history_ignore();
+        // just for convenience and to make the code look a bit cleaner
+        let ignore = &self.ignore_setting.flags;
+        let regexes = &self.ignore_setting.regexes;
 
         // without the second check the command which sets the environment variable would also be
         // ignored. However, this behavior might not be wanted.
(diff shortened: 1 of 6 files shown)


mmstick:
I'll wait for the TODO's to give you some time to implement them and perfect this feature. I'm not sure why the regex's wouldn't work though, if you ensure that they're set with single quotes (which ignores all parsing rules). IE:

```
echo one two three | awk '{print $2}'
```

mmstick:
Also, maybe we should just make the `HISTORY_IGNORE` array a sort of facade, whereby a match to that key doesn't search our hashmap or arrays, but points directly to a structure which we can update?

BafDyce:
Uh, that's my mistake then. I always used double quotes. That probably explains the issues I had.


Is there already a variable that uses such a behavior in the code base? If not do you have a suggestion on how I could implement this (just a snippet of (pseudo) code would be enough)

Either way, I'll try to resolve the Todos asap but I cant promise much in terms of when it will be finished.

mmstick:
Check out get_var on master. There new SWD and MWD variables are doing that.

BafDyce:
I could not use the same logic as it is used for the SWD and MWD variables. (These live just in the context of the Variables struct and we need to access the history).

I therefore went another way:
Add a new member to the Shell struct which contains the patterns (+ compiled regexes) that specify which commands should be ignored from the history. Whenever an array variable is asigned some value we check whether it was `HISTORY_IGNORE` that was changed. If yes, then we update the shells ignore patterns. Compilation of all regexes also happens **only** at this time.

> Also, maybe we should just make the HISTORY_IGNORE array a sort of facade, whereby a match to that key doesn't search our hashmap of arrays, but points directly to a structure which we can update?

I didn't do that (`HISTORY_IGNORE` is still a normal array variable) so that one can still use it like an array in shell scripts, etc.

> I'm not sure why the regex's wouldn't work though, if you ensure that they're set with single quotes (which ignores all parsing rules)

Some testing confirmed this! When using single quotes everything works as expected.

*I'll also update the original post to reflect the new changes*

BafDyce:
Rebased the branch (+ `git push -f` !) to resolve the merge conflicts caused by fab14ab. I also verified that the feature still works as expected.

mmstick:
Sorry about the big change. There shouldn't have been any conflicts between our work. Changes were largely within the assignment parsing, and changes outside of that were strictly to fix compiler errors due to the change in the API for assignment parsing. Assignment parsing should now be much more advanced, and I'll have all the type checking functionality completed later today. Maybe even implement array arithmetic while I'm at it. Something like this:

```ion
let a:int[] = [1 2 3 4]
let a += 1
echo @a
> 2 3 4 5
```

BafDyce:
No problem, it was just a small conflict which was resolved in about 5-10 minutes. Part of my PR adds a small check to the assignment logic, to detect when `HISTORY_IGNORE` is changed and to update the ignore patterns/regular expressions.

I welcome the changes to the assignment parsing API. I was positively surprised to see types coming into a scripting language. Opens up a few questions though:
- Is it mandatory for each assigment to specify a type or is there some sort of default type (I haven't fully looked through the code yet)
- Will casting be supported? For example, will this be possible in the future?
```
# let a:str = "123"
# let b:int = a
# let a += 4
# let b += 4
# echo $a $b
1234 127
```

mmstick:
If no type parameter is defined, then it's type is determined by the value, same as before. Technically, the type is referred to internally as the `Any` type.

```ion
let a = [one two three]
let a = "one"
```

It only checks to see if the type is an array, as that's all that really matters in deciding where we're going to assign this variable. Arrays of course are defined by whether or not they're wrapped within brackets. No brackets = string.

As for what you mentioned regarding casting, the type checking only happens during assignment, and that information is not carried on after assignment. So you can technically create an `a:str` and then later set a new variable with `b:int` using `a`'s value.

If you attempt to use an assignment operator on a value whose parameter isn't defined, it will try to determine if that value can be parsed as a `u64`, followed by `f64`, and then otherwise error.

BafDyce:
Sounds neat :-)

>  it will try to determine if that value can be parsed as a `u64`, followed by `f64`, and then otherwise error.

Do I read this correctly that signed integers (`i64`) are not (yet) implemented/supported? Or are signed values covered by `f64`?

mmstick:
Forgot to think about it, but I'll do that. `u64` -> `i64` -> `f64`.

mmstick:
Okay -- I have type checking completed, for both arrays and strings. And switched to `i64` instead of `u64`. Boolean values will also be normalized to a standard value.

```ion
let a:bool = [1 0 1 1 0 0]
echo @a
> true false true true false false
```

BafDyce:
Rebased again to include latest hotfix.

mmstick:
Do you want me to merge this now?

src/shell/history.rs
```diff
@@ -58,4 +102,89 @@ impl<'a> ShellHistory for Shell<'a> {
...
     }
+
+    fn save_command_in_history(&mut self, command: &str) {
+        if self.should_save_command(command) {
+            // Mark the command in the context history
+            self.set_context_history_from_vars();
+            if let Err(err) = self.context.as_mut().unwrap().history.push(command.into()) {
+                let stderr = io::stderr();
+                let mut stderr = stderr.lock();
+                let _ = writeln!(stderr, "ion: {}", err);
```
mmstick: You could use `eprintln!()` instead to make this simpler. Does the same thing.

src/shell/history.rs
```diff
@@ -58,4 +102,89 @@ impl<'a> ShellHistory for Shell<'a> {
...
+        let regex_prefix = "regex:";
+        for pattern in patterns {
+            match pattern.as_ref() {
+                "all" => flags |= IGNORE_ALL,
+                "no_such_command" => flags |= IGNORE_NO_SUCH_COMMAND,
+                "whitespace" => flags |= IGNORE_WHITESPACE,
+                // The length check is there to just ignore empty regex definitions
+                _ if pattern.starts_with(regex_prefix) && pattern.len() > regex_prefix.len() => {
+                    flags |= IGNORE_BASED_ON_REGEX;
+                    let regex_string = pattern[regex_prefix.len()..].to_owned();
```
mmstick: Maybe only use `to_owned()` when pushing to the `regexes` vector?

src/shell/history.rs
```diff
@@ -58,4 +102,89 @@ impl<'a> ShellHistory for Shell<'a> {
...
+        }
+
+        // Here we allow to also ignore the setting of the environment variable because we assume
+        // the user entered the leading whitespace on purpose.
+        if ignore.contains(IGNORE_WHITESPACE) {
+            if let Some(c) = command.chars().next() {
+                if c.is_whitespace() {
+                    return false;
+                }
+            }
```
mmstick: Can rewrite this block as:

```rust
if command.bytes().next().map_or(false, |b| (b as char).is_whitespace()) {
    return true;
}
```

src/shell/history.rs
```diff
@@ -58,4 +102,89 @@ impl<'a> ShellHistory for Shell<'a> {
...
+
+        if let Some(ref regexes) = *regexes {
+            for regex in regexes {
+                // ignore command when regex is matched but only if it does not contain
+                // "HISTORY_IGNORE", otherwise we would also ignore the command which
+                // sets the variable, which could be annoying.
+                if regex.is_match(command) && !command.contains("HISTORY_IGNORE") {
+                    return false;
+                }
+            }
```
mmstick:
```rust
if regexes.any(|regex| regex.is_match(command) && !command.contains("HISTORY_IGNORE")) {
    return false;
}
```

BafDyce:
Thanks for your feedback! I'll include the suggested changes.

Also, I've the documentation almost finished (added an entire page to the manual, just about the command history). So, I'd like to at least include the documentation into the PR.

src/shell/history.rs
BafDyce: Shouldn't we use `command.chars()` here?

We wouldn't need the cast and nothing breaks if the first character is some multi-byte unicode char.

src/shell/history.rs
BafDyce: Actually, we don't need `to_owned()` at all here. It's a surviver of some earlier version and I forgot to remove it when refactoring.

src/shell/history.rs
BafDyce: This looks so much cleaner! I really should get more familiar with closures and all the fancy functions that are available in the standard lib ;)

src/shell/history.rs
mmstick: Has higher performance to use bytes. Whitespace characters are never larger than a byte, so it's a bit more efficient.

BafDyce:
If you want, you can merge it now.

Unless you also want to have some unit & integration tests included. Decision is upon you ;)

mmstick:
You can add the unit & integration tests later. I'll go ahead and merge this.

BafDyce:
Maybe, we should have removed the "WIP:" from the title prior to the merge :D

mmstick:
As a heads up, I'm writing the documentation for method expansions at the moment.

mmstick:
Documentation for methods is complete. Now documenting the new assignment logic.

mmstick:
And that's now complete.

Issue #416, opened 2017-07-15
Better Builtin Help Documentation


mmstick:
Currently, builtins have very basic one-line short descriptions, but what we need is to provide a complete manpage when help for a specific command is asked for.

sebastianpfluegelmeier:
Could you give an example for one buildin. i would like to enhance the documentation but i don't know what you imagine.

PR #514, opened 2017-08-24
Fix small error in readme

BafDyce:
Removed just some stray words.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ We are providing our manual for Ion in the form of a markdown-based book, which 
 
 - The official page for Ion's manual on Redox's website: https://doc.redox-os.org/ion-manual/
 - Installing the mdbook via our `setup.ion` script and having Ion open an offline copy via `ion-docs`.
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks! Any ideas for what we could document better?

BafDyce:
Right now, the only thing that comes to my mind would be to improve/add documentation of the builtins, as is already mentioned in #416 Maybe even add some documentation of the various methods. (It's a bit strange that these are one of ions most powerful features and yet there's almost no documentation about them)

This includes the "in-shell" documentation (via `help`) as well as the manual.

mmstick:
Reason for no documentation at the moment is because it would take a bit of effort to document each and every method we support right now. It is a high priority to document them though.

BafDyce:
I see that it takes a lot of effort. (Thats why most projects lack documentation :P) When I have finished the feature-implementations I'm currently working on, I can help documenting! I'll come back to you to ask where I should begin once I am ready.

Issue #515, opened 2017-08-24
Add a mechanism to get ions version


BafDyce:
**Expected behavior**:
It would be great if ion would provide some mechanisms to retrieve some version information about the binary. e.g. via `help about` or `ion --version`
This information could include a "fixed" version such as `1.0.5 Alpha` and/or the latest commit hash when the binary was compiled, as well as (optional) the systems' version.

This is especially helpful during development to quickly check if one runs the "new" version or not. 

**Actual behavior**:
Currently, ion does not provide any mechanism to retrieve its version.

**Misc**:
To be honest, I don't know how hard it is to retrieve the latest commits hash at compile time. Another useful addition would be to display additional information such as the current (at compile time) branch and whether there were unstaged changes.

mmstick:
Very easy to get the commit hash, actually. Off the top of my head, we could modify the `build.rs` file to grab the commit and write the hash to a file, then just use an `inline_str!(FILEPATH)` macro to embed that string into the binary.

drosseau:
Seems like this wasn't auto closed because I put the wrong format in the PR. It can be closed now I believe.

PR #516, opened 2017-08-24
Added a --version flag

drosseau:
Adds a flag to display the version and git rev.

```
$ ion --version
ion 1.0.5 (x86_64-unknown-linux-gnu)
rev d1d02727024f98b199490122604bc8a7712c2fe4
```

# Issues

`include!(concat!(env!("OUT_DIR"), "/version_string"))` might eventually be an issue if `/` isn't the default path separator though. You can't use `Path::MAIN_SEPARATOR` there.

First this tries to get the revision from `git rev-parse master` and then falls back to trying the `.git/refs/heads/master` file if the command fails. That kinda locks in the build to master branch, but I think that is OK?

(Sorry about the messy commits..)

This would close issue #515

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -4,7 +4,7 @@
 
 **Actual behavior**: [describe the actual behavior, which is presented through the repro.].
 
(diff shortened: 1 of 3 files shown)


mmstick:
Thanks! This is awesome.

BafDyce:
Awesome! Thanks!

PR #522, opened 2017-09-08
refactor `match` statements

nivkner:
The original problem I had was that when trying to compile `a6e0e86` was that I got errors such as:
```
error: expected one of `,`, `.`, `?`, or `}`, found `Err`
  --> src/shell/assignments.rs:70:33
   |
69 |                                 }
   |                                  - expected one of `,`, `.`, `?`, or `}` here
70 |                                 Err(why) => {
   |                                 ^^^ unexpected token
```

The cause of it might have been the inner `match`'s `{}` being confused for a multi-line `match` arm.

I have no idea how that slipped by, since its a syntax error. I deleted and cloned the repository again just in case; still got the same error.

Instead of simply adding `,` at the end, I decided refactor the methods to avoid such problems altogether.

This PR addresses readability issues, such as unnecessary nested `match`s,  variable names being used multiple times for different items, or functions being used outside of the match statement as though independent of the outcome,  despite only happening in the first arm's case.

### DIFF
--- a/src/shell/assignments.rs
+++ b/src/shell/assignments.rs
@@ -62,38 +62,31 @@ impl<'a> VariableStore for Shell<'a> {
                 for action in assignment_actions {
                     match action {
                         Ok(Action::UpdateArray(key, Operator::Equal, expression)) => {
-                            let values = match value_check(self, &expression, key.kind) {
-                                Ok(values) => match values {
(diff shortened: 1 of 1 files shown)


mmstick:
Are you using the nightly toolchain? Rustfmt formats code in a way that's incompatible with stable at the moment.

PR #523, opened 2017-09-09
distinguish between environment and local variables

nivkner:
In the history sub-module, only the local variables are used to determine things like the history size, whereas the documentation keeps referring to them as environment variables, which might confuse users who might attempt to export `HISTORY_SIZE` from bash for example, only to find out the changes didn't register.


BafDyce:
Whoops, my mistake! Thanks for fixing!

PR #536, opened 2017-09-20
Fix Redox watch_foreground

jackpot51:
This makes the Redox `watch_foreground` function look much more like the Unix one. It also fixes a bug with: `something | less`


jackpot51:
Thanks @mmstick !

PR #537, opened 2017-09-22
use `c_char` for plugin arguments instead of `i8`

nivkner:
Treating `c_char` and `i8` as interchangeable makes program less portable since these types are defined differently per-platform.

Use the `c_char` alias whenever applicable instead of `i8`.  

Find and fix other such problems with types used without aliases and maybe switch to higher-lever wrapper libraries to prevent such issues altogether.


nivkner:
I finished converting the rest of the `i8` to `c_char`, which rust wasn't complaining about since `libloading` is generic over imported symbols.

Issue #545, opened 2017-10-06
Multiple Redirection Support


mmstick:
It's currently only possible to perform one redirection at a time. Adding support for multiple redirections would be very useful in a number of cases, and should be somewhat easy to implement. In example, redirecting standard output to /dev/null, but redirecting standard error to a log.

```sh
cmd args... > /dev/null ^> log
```

chrisvittal:
I'm willing to tackle this.
Open question, what should happen in cases where both argments are the same (like `cmd args... > file ^> file`)?

mmstick:
In the long term, it may be ideal to enable sending identical outputs to multiple places, so that the above command would send a copy of `stdout` to `file1`, and a copy of `stdout+stderr` to file2. Reason being that there are use cases for performing redirections like so:

```sh
cmd args... > log | cmd
```

Which writes a copy of the stdout of the first command into a file named log, and sends another copy to the stdin of the second command. Although it's fine if we set the current goal to just enabling multiple redirections, and possibly notifying the scripter that redirection the same output to multiple locations isn't supported yet.

mmstick:
The `memfd_create` syscall may be the way to go for implementing multi-location redirection.

chrisvittal:
I'll look into it. Interestingly, for the following command
```
cat file > newfile1 > newfile2
```
There is discrepancy between shells, bash and dash only populate `newfile2`, zsh has both files as a copies of `file`.

My current design simply adds another redirection. Though we could potentially turn the field in `Pipeline` into a vector of redirections, that then get processed. I will look into doing that.

I guess the idea would be to use `memfd_create` to create a file in memory that we then copy where it needs to go?

mmstick:
Yes, I would guess that the ideal means would be an enum that declares whether there's an output to a single location, or a multi-location redirection, with the multi-location redirection consisting of a vector of locations. `memfd_create` would be invoked to safely store and copy the outputs to multiple locations. The kernel will delete the anonymous file in memory when all FD handles are dropped.

mmstick:
Redox doesn't support the `memfd_create` call at this time, however, so support for Redox would have to be implemented using the traditional means of writing outputs to a file in the tmp directory, and manually deleting that file when done.

chrisvittal:
I was going to ask that, Thanks! I should be able to have something done today. I've replaced the `Option<Redirection>` with a new tri-state enum (`None`, `Single`, `Multiple`). I'll keep hacking away on this.

[...]

PR #548, opened 2017-10-09
WIP: Add multi stream redirection

chrisvittal:
Add enum `RedirectKind` to consider three distinct cases of redirection.

    1. No redirection [RedirectKind::None]
    2. Single redirection [RedirectKind::Single(Redirection)]
    3. Multiple redirections [RedirectKind::Multiple(Vec<Redirection>)]

The intention is to have the Single case work as it always has, and to
use tempfiles of some kind to handle the the multiple output case.

What is still left to do:
* Turning the memory mapped files into all the real files we need and making sure the following cases are handled.
    ```
    cmd args... > newfile >> appendedfile # no output (or error) redirection
    cmd args... >> output &> out-and-err # an output and a both out+err redirection
    ```
* Redox support (there is currently only a stub)
* (Possibly) reorganize it so that all platform specific code is internal to platform specific files

**Fixes**: #545

**State**: WIP

### DIFF
--- a/src/parser/pipelines/collector.rs
+++ b/src/parser/pipelines/collector.rs
@@ -3,7 +3,7 @@
 use std::collections::HashSet;
 use std::iter::Peekable;
 
-use super::{Input, Pipeline, RedirectFrom, Redirection};
+use super::{Input, Pipeline, RedirectFrom, Redirection, RedirectKind};
 use shell::{Job, JobKind};
 use types::*;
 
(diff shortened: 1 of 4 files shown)


mmstick:
How's the progress going with this? Having any difficulties?

chrisvittal:
Sorry for the delay. I've been basically completely retooling this. The retool is also not quite finished yet. I'm going to open a PR from that tree so we get a diff and can discuss sortfalls, alternatives, how to make the new one work (etc.)

PR #550, opened 2017-10-09
Change to "is_array" to properly return false when multiple arrays are present

Kvikal:
**Problem**: Function "is_array" returned true on input containing multiple arrays. According to part of "is_array_" test (was commented out before issue was fixed), this should not be happening.

**Solution**: Added additional check that detects the end of topmost bracket and returns false if there are characters following the end of the bracket.

- Enabled part of "is_array_" test that was previously commented out
- Added enum "IsArrayHelper"
- Altered "is_array" function

**State**: ready

**Other**:  I am not very experienced with Rust, so this might not be ideal way to implement the check for multiple brackets.

### DIFF
--- a/src/parser/assignments/checker.rs
+++ b/src/parser/assignments/checker.rs
@@ -4,12 +4,11 @@ use super::super::expand_string;
 
 use std::iter::Iterator;
 
-#[derive(PartialEq,Clone)]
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

PR #553, opened 2017-10-16
Implement mutiple redirection.

chrisvittal:
Add multiple redirections, that is, support for things like:

```
cmd <file1 <file2 ^>> err-log > output | cmd2
```
In current ion this doesn't work.

This change at current would be breaking as the statement:
```
tr 'x' 'y' | tr 'y' 'z' < file > file2
```
redirects `file` to the first `tr` command. With this change, we would redirect it to the second.

## Overview
We treat a multiple redirection as it's own job. The only difference between these jobs and normal jobs is that they only read from their source(s) and then write to their sink(s).

### Parsing
Alter `Pipline` contain a vector of `PipeItem`s. A `PipeItem` consists of a `Job`, a `Vec<Redirection>` and a `Vec<Input>`. The above statement would be parsed into two items, `cmd` together with `[file1, file2]` as input and `[err-log, output]` as outputs, and `cmd2` with no `Input`s or `Redirection`s

### Command Generation and Redirection Setup
Added two two new variants of `RefinedJob`, `Cat` and `Tee`, `Cat` for concatenating multiple inputs, `Tee` for splitting multiple outputs. Alter the `generate_commands` function to return a `Vec<(RefinedJob, JobKind, Vec<Redirection>, Vec<Input>)>`, then condense `redirect_input` and `redirect_output` into one function that takes ownership and returns a new `Vec<(RefinedJob, JobKind)>`. It is here that we create the `Tee` and `Cat` jobs that we need and put them in the right spot in the pipeline.

### Execution
Add `exec_multi_in` and `exec_multi_out` to handle the multiple redirections. These are fairly straightforward, they just read their sources and write to their sinks. I didn't add any extra paths to `exec_job` as the new `RefinedJob` kinds are always part of a pipeline.

### DIFF
--- a/src/parser/pipelines/collector.rs
+++ b/src/parser/pipelines/collector.rs
@@ -372,7 +372,7 @@ impl<'a> Collector<'a> {
 
 #[cfg(test)]
 mod tests {
-    use parser::pipelines::{Input, Pipeline, RedirectFrom, Redirection};
+    use parser::pipelines::{Input, PipeItem, Pipeline, RedirectFrom, Redirection};
     use parser::statement::parse;
     use shell::{Job, JobKind};
     use shell::flow_control::Statement;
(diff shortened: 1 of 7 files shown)


jackpot51:
Sweet!

mmstick:
I've been thinking that the whole job control part likely needs to be carefully reviewed to ensure that pipes are getting closed on both ends properly. Could be why you are having issues.

mmstick:
If you need help, I can try to look into why your pipes aren't getting closed.

chrisvittal:
Sorry for taking so long; I think I just needed to sit on it and mull it over. The solution is closing the pipes in the parent process. I'll need to find some way of doing that effectively for 

I would appreciate help testing it out, the solution appears to be closing the pipes in the appropriate parent job. I need to find a way to work with `RefinedJob::External`, but there is probably something I can do.

Thanks for the offer. I'll have something pushed later tonight.

chrisvittal:
So here we are. This needs more testing. Especially with long running jobs and other such things. It seems portable, I don't think I used linux vs redox specific code, so hopefully it shoud be portable. Let me know if I did, I'd be interested in fixing it up.

mmstick:
Thanks! This is a huge help to Ion.

mmstick:
Thanks again. This new feature is incredible!

PR #575, opened 2017-11-04
Fix string comparisons in documentation

xTibor:
The builtin test command expects `=` for string comparisons, not `-eq`.


mmstick:
Thanks!

Issue #556, opened 2017-10-21
$escape() and $unescape() methods


mmstick:
Essentially, Bash allows one to convert ASCII sequences into their corresponding codes if that string is enclosed within single quotes that follow the `$`.

```bash
echo $'this\nstring\thas\tspecial\ncodes'
```

```
this
string	has	special
codes
```

We should support this with an `$unescape()` method, like so:

```sh
echo $unescape('this\nstring\thas\tspecial\ncodes')
```

As well as a corresponding `$escape()` method that will do the opposite. We could likely re-use logic from the `echo` builtin to do this.

AgustinCB:
Can I take this task? It will probably take some days, since I need to get used to the code. But I think I understand the parts that are involved and how to do it.

mmstick:
Sure. You would implement this [here](https://github.com/redox-os/ion/blob/master/src/parser/shell_expand/words/methods/strings.rs#L84), with the rest of the string methods.

PR #579, opened 2017-11-07
Escape unescape

AgustinCB:
I removed some things from the template that looked like would contain duplicated answers or no answer at all.

**Problem**: Need a way to unescape strings, similar to $"\n".

**Solution**: I created two new methods, escape and unescape.

**Drawbacks**:
- The logic about how to unescape is repeated in the method and in the builtin echo. I wonder if it's possible to dedup that.
- No tests. I didn't know how to write them

**Fixes**: #556

**State**: Ready. I guess. More like: Waiting for feedback :).

### DIFF
--- a/src/parser/shell_expand/words/methods/strings.rs
+++ b/src/parser/shell_expand/words/methods/strings.rs
@@ -257,14 +257,13 @@ impl<'a> StringMethod<'a> {
             },
             "escape" => {
                 let word = if let Some(value) = expand.variable(variable, false) {
-                    Some(value)
+                    value
                 } else if is_expression(variable) {
-                    Some(expand_string(variable, expand, false).join(" "))
+                    expand_string(variable, expand, false).join(" ")
                 } else {
-                    None
+                    return;
                 };
                 let out: Vec<String> = word
-                    .unwrap_or(String::from(""))
                     .chars()
                     .map(|c| c.escape_default().to_string())
                     .collect();
--- a/src/parser/shell_expand/words/methods/strings.rs
+++ b/src/parser/shell_expand/words/methods/strings.rs
@@ -184,6 +184,80 @@ impl<'a> StringMethod<'a> {
                 };
                 output.push_str(&out.unwrap_or(0).to_string());
             },
+            "unescape" => {
+                let word = if let Some(value) = expand.variable(variable, false) {
+                    Some(value)
+                } else if is_expression(variable) {
+                    Some(expand_string(variable, expand, false).join(" "))
+                } else {
+                    None
+                };
+                let mut out = String::from("");
+                let mut check = false;
+                let input = word
+                    .unwrap_or(String::from(""))
+                    .to_string();
+                for c in input.chars() {
+                    match c {
+                        '\\' if check => {
+                            out.push(c);
+                            check = false;
+                        }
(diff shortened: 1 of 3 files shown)


src/parser/shell_expand/words/methods/strings.rs
```diff
@@ -183,7 +183,96 @@ impl<'a> StringMethod<'a> {
...
+            },
+            "unescape" => {
+                let word = if let Some(value) = expand.variable(variable, false) {
+                    Some(value)
+                } else if is_expression(variable) {
+                    Some(expand_string(variable, expand, false).join(" "))
+                } else {
+                    None
+                };
+                let mut out = String::from("");
```
mmstick: I would definitely set a size here to prevent future reallocations.

```rust
let mut out = String::with_capacity(input.len());
```

src/parser/shell_expand/words/methods/strings.rs
```diff
@@ -183,7 +183,96 @@ impl<'a> StringMethod<'a> {
...
                 output.push_str(&out.unwrap_or(0).to_string());
-            }
+            },
+            "unescape" => {
+                let word = if let Some(value) = expand.variable(variable, false) {
+                    Some(value)
+                } else if is_expression(variable) {
+                    Some(expand_string(variable, expand, false).join(" "))
+                } else {
+                    None
```
mmstick: Maybe simply return if the value is `None`? Would avoid the need for `Option` here.

mmstick:
You could deduplicate the logic by creating a function take accepts a closure. Something like:

```rust
fn unescape<F: Fn(u8)>(byte: u8, func: F);
```

mmstick:
Anyway, thanks for the implementation! I shall merge this tomorrow morning.

AgustinCB:
I addressed some of the comments and left some structure to do the code sharing thingy tomorrow.

Thanks for the feedback!

src/parser/shell_expand/words/methods/strings.rs
```diff
@@ -183,7 +183,92 @@ impl<'a> StringMethod<'a> {
...
+                } else if is_expression(variable) {
+                    expand_string(variable, expand, false).join(" ")
+                } else {
+                    return;
+                };
+                let out: Vec<String> = word
+                    .chars()
+                    .map(|c| c.escape_default().to_string())
+                    .collect();
+                output.push_str(&out.join(""));
```
mmstick: We could eliminate some heap allocations here by writing this as:

```rust
#[feature(str_escape)]
output.push_str(&word.escape_default());
```

Although we may want to escape characters ourselves, so that we can escape special characters relevant to system shells like Ion. Something efficient like:

```rust
fn escape_string(output: &mut String, input: &str);
```

Then escaping all special characters in the 0-127 ASCII range, except for _, :, and path::MAIN_SEPARATOR.

AgustinCB: Yeah, I was wondering if I should create an specific escape logic here. It makes sense. Will make the change after work!

mmstick:
Is this ready to merge?

AgustinCB:
If you have no further feedback, yes!

mmstick:
Thanks!

PR #583, opened 2017-11-07
Documented Builtins

KaKnife:
Documented Builtin utilites


mmstick:
Thanks! There's also `drop -a VARIABLE` for dropping array variables. But I'll add that with a followup commit.

Issue #586, opened 2017-11-13
The test builtin fails silently on negative values


xTibor:
This might trip some users coming from GNU. The GNU implementation supports comparisons with signed values while Ion's `test` only supports unsigned values and fails silently on signed ones.

**Reproduction**:
```
test "-100" -eq "200" && echo true || echo false
test "-100" -ne "200" && echo true || echo false
test "-100" -gt "200" && echo true || echo false
test "-100" -lt "200" && echo true || echo false
```

**Expected behavior**:
```
false
true
false
true
```

**Actual behavior**:
```
true
true
true
true
```

**Build information**:
`rustc -V`: `rustc 1.23.0-nightly (d6b06c63a 2017-11-09)`
`git rev-parse HEAD`: 2f176ebe1010b4ebd9e2a9031ad354e598cdee9c
System: KDE neon 5.11 (Ubuntu 16.04 LTS)

mmstick:
Sounds like a simple fix.

mmstick:
I wonder if we should parse values with the `decimal` crate? So that we can parse them as 128-bit decimals, which thereby would support both negative values, and values with decimal places?

PR #590, opened 2017-11-16
Fix test with negative numbers

AgustinCB:
**Problem**: the `test` builtin is falsely passing when one of the numbers is negative.

**Solution**: Two problems were addressed:
- It was assuming that negative numbers were flag strings: Solved by checking that the second character is alphabetic.
- It was parsing to `usize`. Solved by parsing to `isize`.

Other than what commented in the solution, added tests for this cases.

**TODOs**: In the issue was commented the possibility of using `decimal` and parsing to `d128`. I didn't feel comfortable adding that overhead without a deeper discussion, so I just made the minimal fix to solve the problem.

Also... is `is_alphabetic` a good match? I felt tempted to use `is_ascii_alphabetic` instead, but it's still experimental and it wasn't considered a `Pattern`, so the compiler complained. `is_alphabetical` cover more cases than we do right now, so maybe is worth to use our own `is_ascii_alphabetic`?

**Fixes**: #586

**State**: Ready

### DIFF
--- a/src/builtins/test.rs
+++ b/src/builtins/test.rs
@@ -126,7 +126,7 @@ fn evaluate_arguments<W: io::Write>(arguments: &[&str], buffer: &mut W) -> Resul
             buffer.flush().map_err(|x| x.description().to_owned())?;
             Ok(true)
         }
-        Some(&s) if s.starts_with("-") => {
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

PR #593, opened 2017-11-16
Build for Redox on Travis

ids1024:
I've been meaning to do this, but only recently took the time to install a Debian chroot and figure out an issue that was interfering with it (https://github.com/redox-os/libc/pull/43).

I've sent a similar PR to uutils: https://github.com/uutils/coreutils/pull/1095


mmstick:
This will be very helpful.

Issue #563, opened 2017-10-30
Perform Process Expansions Output Handling Without Allocating


mmstick:
When process expansions aren't quoted, we split the output by whitespace and then join them together with spaces. We need a better solution that doesn't involve allocations. Example code:

```rust
// TODO: Complete this so that we don't need any heap allocations.
// All that we need is to shift bytes to the left when extra spaces are found,
// and to disallow whitespace at the beginning and end of the string.

unsafe {
    let bytes: &mut [u8] = output.as_bytes_mut();
    bytes.iter_mut()
        .filter(|b| **b == b'\n' || **b == b'\t')
        .for_each(|b| *b = b' ');
    slice(current, str::from_utf8_unchecked(&bytes), selection)
}
```

PR #596, opened 2017-11-18
Less heap

AgustinCB:
**Problem**: expand_process have more allocations than necessary.

**Solution**: Mutate the output in place instead of splitting and then joining.

- Added a test for unquoted strings for the expand_process function.

**Drawbacks**: It uses `unsafe`. It's possible to do something "similar" (although not that conservative) without going `unsafe`. In fact that was my first iteration and that solution is in the first and second commit. If you prefer that one instead of the `unsafe` alternative, let me know and I'll revert the last commit.

**TODOs**: Add test for quoted strings! (although out of the scope of this pr).

**Fixes**: #563

**State**: Ready

### DIFF
--- a/src/parser/shell_expand/mod.rs
+++ b/src/parser/shell_expand/mod.rs
@@ -50,15 +50,24 @@ fn expand_process<E: Expander>(
             };
             slice(current, output, selection)
         } else {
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks! There's always room for some tests in any PR.

PR #603, opened 2017-11-25
Add builtin status

Sag0Sag0:
**New builtin, status.**

Accepts three arguments.
**-l** returns true if the shell is a login shell.
**-i** returns true if the shell is interactive.
**-f** prints the filename of the currently running shell script or if there is none prints stdio.

**Potential problem**
`status` works out if the current shell is a login shell by checking that `arg[0][0] == '-'`, that is the convention on unix based systems. However I'm unsure if redox does this.


mmstick:
Thanks!

Issue #570, opened 2017-11-03
Ion Manual Improvements


mmstick:
This issue will highlight all the critical issues with the Ion manual. Feel free to add to this list in the comments below.

# Undocumented Features

- [ ] Functions
    - [ ] Optionally type-checked parameters
    - [ ] Piping functions
- [ ] Match syntax
- [ ] Word Designators
- [ ] Namespaces
    - [ ] Color namespace & 24-bit colors
- [ ] Builtin Commands
- [ ] Prompt Function & Prompt Variable
- [ ] Default Variables
- [ ] Signal Handling & Job Control
- [ ] Plugins
    - [ ] String method plugins
    - [ ] String namespace plugins
- [ ] Misc Rules
    - [ ] How quoting rules work with string variables and string command expansions

AgustinCB:
Where can I read samples about this functionalities? I wouldn't mind tackling this task.

mmstick:
@AgustinCB The source code is basically the only reference for the undocumented parts of the manual. If there's a key area that you would like elaborated, I can explain it.

Namespaces, for example, are accessed via string braces (`${namespace::item}`). I've documented the full capabilities of the namespace [on my Rust-written blog here](https://mmstick.tk/post/7A3).

AgustinCB:
Ok. I will take as a learning experience then and go docjmenting those bits as I read more abou the code. Will probably go contributing with small prs covering some of the topics as I get familiar with them.

Thanks!

PR #605, opened 2017-11-25
Test and refactor of array methods

AgustinCB:
**Problem**: I noticed that there's no documentation for the method `split_at`. To write sensitive documentation, I like to look at the tests. For the array methods, there aren't.

**Solution**: Because there weren't, I decided to add them. As with the string methods, tests, I'll use them to update The Book later.

- e2e tests for array methods.
- Make `run_examples.sh` tell you the difference on fail.
- Unit tests for the `ArrayMethod` struct.
- Small refactor to reduce code duplication.
- Small refactor on `graphemes` to avoid cloning the structure.

**TODOs**: I have the feeling that the `split` function can slightly optimized. I'm not sure if that's worth though, I'll have to think on a way to benchmark it first.

**Fixes**: I'm very slowing advancing in #570.

**State**: Ready for review.

### DIFF
--- a/examples/array_methods.ion
+++ b/examples/array_methods.ion
@@ -1,0 +1,5 @@
+echo @split("onetwoone", "two")
+echo @split_at("onetwoone", "3")
+echo @graphemes("onetwo", "3")
+echo @bytes("onetwo")
+echo @chars("onetwo")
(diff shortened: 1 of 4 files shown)


mmstick:
Looks good to me. When you are ready to merge, I will do so.

AgustinCB:
Whenever you want!

mmstick:
It is done.

PR #612, opened 2017-11-28
Update the disown command

dlrobertson:
**Problem**:

 - disown panics if
   - No arguments are provided to the command
   - No jobspec's are provided to the -r or -h options
 - disown does not have a help option

**Solution**:

Add a few extra checks in the implementation of `disown` and add
`disown::print_help`.

**Other:**

I hit this a few times as I was learning how to use disown.


mmstick:
Thanks!

PR #615, opened 2017-11-29
Add functions to manual

AgustinCB:
My English can be very poor, so feel free to give me feedback on that.


src/parser/statement/functions.rs
```diff
@@ -14,14 +14,7 @@ pub(crate) fn parse_function<'a>(arg: &'a str) -> (KeyIterator<'a>, Option<&'a s
...
 pub(crate) fn collect_arguments<'a>(args: KeyIterator<'a>) -> Result<Vec<KeyBuf>, TypeError<'a>> {
-    // NOTE: Seems to be some kind of issue with Rust's compiler accepting this:
-    //     Ok(args.map(|a| a.map(Into::into)?).collect::<Vec<_>>())
-    // Seems to think that `a` is a `KeyBuf` when it's actually a `Result<Key, _>`.
-    let mut output = Vec::new();
-    for arg in args {
-        output.push(arg?.into());
-    }
-    Ok(output)
+    args.map(|a| a.map(Into::into)).collect()
```
AgustinCB: What happened here:

```
Ok(args.map(|a| a.map(Into::into)?).collect::<Vec<_>>())
```

Wouldn't work because it's sending mixed signals to the compiler. By placing this as the return and passing the type in collect, `rustc` is assuming that the type of the elements in the iterator is `KeyBuf` (because that's what the function returns). So it's expecting that the closure in the first map returns a `KeyBuf`.

The "keyword" `?` works only in functions that returns `Result<T, E>`. So when you use it inside the closure (it doesn't apply at a global level, in the function, but in the closure), it then thinks that the closure is returning a `Result`. Because `Result<Key, _>` isn't the same as `KeyBuf`, it complains.

This approach will return an iterator of `Return<KeyBuf, TypeError<'a>`. When saying collect on this kind of iterators with mixed structure, it will do the same as `sequence` in Haskell (http://hackage.haskell.org/package/base-4.10.0.0/docs/Prelude.html#v:sequence): Flip the order of the structures. So it'd collect into `Vec<Result<KeyBuf, TypeError<'a>>>` and then flip it to `Result<Vec<KeyBuf>, TypeError<'a>>`, returning `Ok(_)` when all the elements in the iterator are `Ok` and `Err` when there's at least one `Err` in the iterator. The content of that `Err` would be the content of the first `Err` in the iterator.

mmstick:
Thanks!

PR #620, opened 2017-12-01
Fix "and/or followed by let results in command not found issue"

Sag0Sag0:
Fixed by doing something very similar to evaling all arguments and/or receives. This doesn't seem to cause any problems and fixes the issue.


mmstick:
I don't think the `custom_eval` part is required.

mmstick:
Am really confused. The old behavior that I had before caused serious issues because the `on_command` method is re-parsing the expanded form of the original inputs, which therefore re-runs the StatementSplitter and gets into trouble. Yet now, that exact code is working?

```rust
fn builtin_and(args: &[&str], shell: &mut Shell) -> i32 {
    match shell.previous_status {
        SUCCESS => {
            shell.on_command(&args[1..].join(" "));
            shell.previous_status
        }
        _ => shell.previous_status,
    }
}

fn builtin_or(args: &[&str], shell: &mut Shell) -> i32 {
    match shell.previous_status {
        FAILURE => {
            shell.on_command(&args[1..].join(" "));
            shell.previous_status
        }
        _ => shell.previous_status,
    }
}
```

Caused

Sag0Sag0:
Deleting it.

mmstick:
I was looking at this wrong. This causes the same issues as before.

```sh
mmstick:~/S/ion# true; and $(rustc --version | grep nightly)
ion: syntax error: '(' at position 22 is out of place
```

mmstick:
So, I've gone ahead and reverted the change.

Sag0Sag0:
Good.

mmstick:
Ideally, and/or/not should probably be keywords rather than builtins.

mmstick:
The time keyword is a good example of how they should be similarly implemented as keywords.

PR #624, opened 2017-12-03
Allow builtins in aliases

### LINKED ISSUE #622: Calling cd (a builtin) from an alias
In zsh I often create this alias to **u** to jump to my parent directory:

```zsh
alias u="cd .."
```

I've tried a few variants in ion, but I can't get it to work. My guess is that alias executes a subprocess that has state independent of the parent process (the current shell).

#409 has `exec` on the TODO list. Is exec the blocker to doing something like this in ion? A related design question might be: after exec is implemented, should all aliases be `exec`'d? E.g., aliases would be more like a macro that replaces characters and can modify the running shell's state.

anxiousmodernman:
During a pipeline execution, aliases are expanded and split into an
Array. The first item in this array should be either a builtin or a
command. If it is a builtin, assign the Job.builtin fn pointer field.
Otherwise, assume the alias refers to a command. Fixes #622

### DIFF
--- a/src/shell/mod.rs
+++ b/src/shell/mod.rs
@@ -240,7 +240,11 @@ impl<'a> Shell {
                         .map(String::from)
                         .chain(pipeline.items[job_no].job.args.drain().skip(1))
                         .collect::<Array>();
-                    pipeline.items[job_no].job.command = new_args[0].clone().into();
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

PR #629, opened 2017-12-05
Add ability for which builtin to process multiple arguments

Sag0Sag0:


src/builtins/mod.rs
```diff
@@ -585,34 +585,29 @@ fn builtin_which(args: &[&str], shell: &mut Shell) -> i32 {
...
-        println!("{}: function", command);
-        SUCCESS
-    } else {
-        for path in env::var("PATH").unwrap_or("/bin".to_string()).split(sys::PATH_SEPARATOR) {
-            let executable = Path::new(path).join(command);
-            if executable.is_file() {
-                println!("{}", executable.display());
-                return SUCCESS;
+    let mut result = SUCCESS;
+    'outer: for i in 1..args.len() {
```
mmstick: Would be more efficient to write this as:

```rust
`outer: for arg in &args[1..] { }
```

src/builtins/mod.rs
```diff
@@ -585,34 +585,29 @@ fn builtin_which(args: &[&str], shell: &mut Shell) -> i32 {
...
+        let command = args[i];
+        if let Some(alias) = shell.variables.aliases.get(command) {
+            println!("{}: alias to {}", command, alias);
+            continue;
+        } else if shell.builtins.contains_key(command) {
+            println!("{}: built-in shell command", command);
+            continue;
+        } else if shell.functions.contains_key(command) {
```
mmstick: Should functions come before builtins?

Sag0Sag0: That was how it was before I edited it.

src/builtins/mod.rs
Sag0Sag0: Yes, however that causes an error.

src/builtins/mod.rs
mmstick: May want to change it while you're at it.

Sag0Sag0: Sure.

src/builtins/mod.rs
Sag0Sag0: Just worked out the error was being caused by something else. Implementing it now.

src/builtins/mod.rs
mmstick: I believe you could also just write this as so:

```rust
'outer: for &command in &args[1..] { }
```

And then not have to manually dereference it,

Sag0Sag0: Thanks, sorry about the mistakes, I only recently began learning rust.

mmstick: It's fine. This is part of the purpose of submitting and reviewing pull requests -- pointing out tricks that you may not be aware of yet.

mmstick:
Thanks!

logankaser:
Does this mean read can have a prompt now? That would be very nice.

PR #630, opened 2017-12-05
Command not found hook

### LINKED ISSUE #567: Command Not Found Hook
Ion should do as other shells and offer a means of executing a custom handler in the event of a command not being found. Common use case is for Linux distributions to do a search of their software repository for a package that contains that command.

As for how we do it, I'd think that having a variable define the command to execute would be a good idea, and failing to find that, a function. Something like `let COMMAND_NOT_FOUND = pkgfile`, and if a command isn't found, simply execute `pkgfile <command>` in it's place.

jD91mZM2:
Fixes #567

### DIFF
--- a/src/shell/pipe_exec/command_not_found.rs
+++ b/src/shell/pipe_exec/command_not_found.rs
@@ -1,4 +1,4 @@
-use super::super::{Capture, Function, FunctionError, Shell};
+use super::super::{Capture, Function, Shell};
 use std::process;
 use sys;
 
@@ -10,8 +10,8 @@ pub(crate) fn command_not_found(shell: &mut Shell, command: &str) -> bool {
 
     if let Err(err) = shell.fork(Capture::None, |child| {
         let result = unsafe { function.read() }.execute(child, &["ion", command]);
-        if let Err(FunctionError::InvalidArgumentCount) = result {
-            eprintln!("ion: COMMAND_NOT_FOUND function takes an invalid number of arguments: must be exactly one");
+        if let Err(err) = result {
+            eprintln!("ion: COMMAND_NOT_FOUND function call: {}", err);
         }
     }) {
         eprintln!("ion: fork error: {}", err);
--- a/src/shell/pipe_exec/command_not_found.rs
+++ b/src/shell/pipe_exec/command_not_found.rs
@@ -1,32 +1,24 @@
-use super::super::{Capture, Function, Shell};
-use parser::shell_expand::expand_string;
-use std::io::Read;
+use super::super::{Capture, Function, FunctionError, Shell};
 use std::process;
 use sys;
 
-pub(crate) fn command_not_found(shell: &mut Shell, command: &str) -> Option<String> {
-    let function = shell.functions.get("COMMAND_NOT_FOUND")? as *const Function;
+pub(crate) fn command_not_found(shell: &mut Shell, command: &str) -> bool {
+    let function = match shell.functions.get("COMMAND_NOT_FOUND") {
+        Some(func) => func as *const Function,
+        None => return false
+    };
 
-    let mut output = None;
-
-    match shell.fork(Capture::Stdout, |child| unsafe {
-        let _ = function.read().execute(child, &["ion", command]);
-    }) {
-        Ok(result) => {
-            let mut string = String::new();
-            match result.stdout.unwrap().read_to_string(&mut string) {
-                Ok(_) => output = Some(string),
-                Err(err) => {
-                    eprintln!("ion: error reading stdout of child: {}", err);
-                }
-            }
-        },
-        Err(err) => {
-            eprintln!("ion: fork error: {}", err);
+    if let Err(err) = shell.fork(Capture::None, |child| {
+        let result = unsafe { function.read() }.execute(child, &["ion", command]);
+        if let Err(FunctionError::InvalidArgumentCount) = result {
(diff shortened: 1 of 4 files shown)


src/shell/pipe_exec/command_not_found.rs
```diff
@@ -0,0 +1,35 @@
...
+use std::process;
+use sys;
+
+pub(crate) fn command_not_found(shell: &mut Shell, command: &str) -> Option<String> {
+    let function = match shell.functions.get("COMMAND_NOT_FOUND") {
+        Some(func) => func as *const Function,
+        None => return None // TODO: Use ? on Option whenever we drop support for older rust versions
+    };
+
+    let mut output = None;
```
jD91mZM2: Is capturing output only to print it out later even necessary? I don't know, you tell me.

src/shell/pipe_exec/command_not_found.rs
```diff
@@ -0,0 +1,35 @@
...
+
+    match shell.fork(Capture::Stdout, |child| unsafe {
+        let _ = function.read().execute(child, &["ion", command]);
```
jD91mZM2: What happens when a function takes the wrong number of arguments? Currently it looks like it just does nothing, but maybe that's not optimal. Should I look into this?

src/shell/pipe_exec/mod.rs
```diff
@@ -495,7 +497,11 @@ impl PipelineExecution for Shell {
                     self.watch_foreground(child.id(), child.id(), move || long, |_| ())
                 }
                 Err(e) => if e.kind() == io::ErrorKind::NotFound {
-                    eprintln!("ion: command not found: {}", short);
+                    if let Some(output) = command_not_found(self, &short) {
```
jD91mZM2: Is `short` the wrong variable to use here? Am I being insecure?

src/shell/pipe_exec/command_not_found.rs
```diff
@@ -0,0 +1,35 @@
+use super::super::{Capture, Function, Shell};
+use parser::shell_expand::expand_string;
+use std::io::Read;
+use std::process;
+use sys;
+
+pub(crate) fn command_not_found(shell: &mut Shell, command: &str) -> Option<String> {
+    let function = match shell.functions.get("COMMAND_NOT_FOUND") {
+        Some(func) => func as *const Function,
+        None => return None // TODO: Use ? on Option whenever we drop support for older rust versions
```
mmstick: You can go ahead and use this feature now. Ion can only be built from nightly versions of Rust.

src/shell/pipe_exec/command_not_found.rs
mmstick: It is not necessary. You don't have to capture anything here.

src/shell/pipe_exec/command_not_found.rs
mmstick: It should print an error that an invalid number of arguments was passed / certain arguments had an invalid type.

src/shell/pipe_exec/mod.rs
mmstick: Could use `&job.command`

mmstick: Then again, I guess `short` is the only option here.

src/shell/pipe_exec/command_not_found.rs
jD91mZM2: (It can also be built by stable currently)
Ah yes! Just wanted to make sure, but forgot to make a comment on it in my self-review

mmstick: We use a number of features that are only available on nightly, and this project actually ensures that cargo selects a nightly version of Rust when compiling, even if your default toolchain is set to the latest stable.

src/shell/pipe_exec/mod.rs
jD91mZM2: Awesome, thanks!

src/shell/pipe_exec/command_not_found.rs
jD91mZM2: Hmm, it doesn't print anything when I change it to `Capture::None` and remove all capture code

src/shell/pipe_exec/command_not_found.rs
jD91mZM2: Wow, that is pretty epic! Didn't know that was possible!

src/shell/pipe_exec/command_not_found.rs
```diff
@@ -1,32 +1,24 @@
...
-                Err(err) => {
-                    eprintln!("ion: error reading stdout of child: {}", err);
-                }
-            }
-        },
-        Err(err) => {
-            eprintln!("ion: fork error: {}", err);
+    if let Err(err) = shell.fork(Capture::None, |child| {
+        let result = unsafe { function.read() }.execute(child, &["ion", command]);
+        if let Err(FunctionError::InvalidArgumentCount) = result {
```
mmstick: May also want to handle `FunctionError::InvalidArgumentType(Primitive, String)`. The `FunctionError` type implements `Display`, so you should be able to just write this as:

```rust
if let Err(why) = result {
    eprintln!("ion: COMMAND_NOT_FOUND function hook: {}", why);
}

jD91mZM2: Didn't realize ion had types. That's pretty cool!

mmstick:
Thanks!

PR #631, opened 2017-12-05
Fix 627

AgustinCB:
We weren't filtering empty strings.

This caused two things: The unwrap in empty strings to fail and to try to compact even on paths with less than two parents.

Added some tests.


mmstick:
Thanks!

Issue #608, opened 2017-11-27
Feature request: Implement the exec command as a builtin


dlrobertson:
**Feature Request**

`ion` should support the `exec` builtin function. Many programs rely on it (e.g. `gdb`).

dlrobertson:
I'll start working on this

PR #634, opened 2017-12-07
Add the exec builtin

dlrobertson:
**Problem**:
The `exec` builtin does not exist

**Solution**:
Implement the `exec` builtin as a simple wrapper around the`execve` syscall.

 - Add the necessary functions to builtin
 - Add the necessary interfaces for the execve syscall
   - Add sys::unix::execve
   - Add sys::redox::execve

**TODOs**:

 - [x] `exec` hits an assertion [here](https://github.com/redox-os/kernel/blob/master/src/arch/x86_64/paging/mapper.rs#L105-L109) when run from the original shell on redox. For example:

```
# Works
echo -e "#!/usr/bin/ion\nexec ion" > ./test.ion
ion ./test.ion
# Hits assertion
exec ion
```
**Resolution**: This occurred due to the context history not being prepped for exit.

**Fixes**:
 - #608

### DIFF
--- a/src/builtins/exec.rs
+++ b/src/builtins/exec.rs
@@ -1,0 +1,38 @@
+use builtins::man_pages::{check_help, MAN_EXEC};
+use shell::Shell;
+use std::error::Error;
+use sys::execve;
+
+/// Executes the givent commmand.
+pub(crate) fn exec(shell: &mut Shell, args: &[&str]) -> Result<(), String> {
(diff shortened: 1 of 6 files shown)


dlrobertson:
Figured out why I was hitting issues on redox. It was because I wasn't correctly prepping the context history for exit. AFAIK this is now ready for review.

mmstick:
Thanks!

PR #635, opened 2017-12-07
Disown using &!

### LINKED ISSUE #539: Syntax for completely launching and disowning a process
### Feature request

I would like to see something similar to `&!` in zsh ([documentation](http://zsh.sourceforge.net/Doc/Release/Jobs-_0026-Signals.html#Jobs-_0026-Signals)).

This allows launching e.g. `firefox &!` and keep it running even after the shell (and the terminal window) is closed.

In addition, I would like to see something like `&!!` (not implemented in ZSH or any other shell I know of) which also detaches `stdin` and `stdout` from the spawned process, as some GUI programs can be noisy. This would be a lot cleaner than having to manually redirecting to `/dev/null`.

jD91mZM2:
Fixes #539, or at least partly.

Warning: Untested! Disowning doesn't seem to do anything, processes aren't closed either way.

### DIFF
--- a/src/parser/pipelines/collector.rs
+++ b/src/parser/pipelines/collector.rs
@@ -281,6 +281,10 @@ impl<'a> Collector<'a> {
                             bytes.next();
                             try_add_item!(JobKind::And);
                         }
+                        Some(&(_, b'!')) => {
+                            bytes.next();
+                            try_add_item!(JobKind::Disown);
+                        }
                         Some(_) | None => {
                             try_add_item!(JobKind::Background);
                         }
@@ -638,6 +642,16 @@ mod tests {
         if let Statement::Pipeline(pipeline) = parse("echo hello world &") {
             let items = pipeline.items;
             assert_eq!(JobKind::Background, items[0].job.kind);
+        } else {
+            assert!(false);
+        }
+    }
+
+    #[test]
+    fn disown_job() {
+        if let Statement::Pipeline(pipeline) = parse("echo hello world&!") {
+            let items = pipeline.items;
+            assert_eq!(JobKind::Disown, items[0].job.kind);
         } else {
             assert!(false);
         }
--- a/src/parser/pipelines/mod.rs
+++ b/src/parser/pipelines/mod.rs
@@ -84,6 +84,7 @@ impl Pipeline {
         self.items.len() > 1 || self.items.iter().any(|it| it.outputs.len() > 0)
             || self.items.iter().any(|it| it.inputs.len() > 0)
             || self.items.last().unwrap().job.kind == JobKind::Background
+            || self.items.last().unwrap().job.kind == JobKind::Disown
     }
 }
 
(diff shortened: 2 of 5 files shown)


src/parser/pipelines/collector.rs
```diff
@@ -281,6 +281,10 @@ impl<'a> Collector<'a> {
                             bytes.next();
                             try_add_item!(JobKind::And);
                         }
+                        Some(&(_, b'!')) => {
+                            bytes.next();
+                            try_add_item!(JobKind::Disown);
```
jD91mZM2: Would it be better to somehow use the existing JobKind with a boolean or something, so it doesn't have to be checked against both Disown and Background every time?

src/parser/pipelines/mod.rs
```diff
@@ -84,6 +84,7 @@ impl Pipeline {
         self.items.len() > 1 || self.items.iter().any(|it| it.outputs.len() > 0)
             || self.items.iter().any(|it| it.inputs.len() > 0)
             || self.items.last().unwrap().job.kind == JobKind::Background
+            || self.items.last().unwrap().job.kind == JobKind::Disown
```
jD91mZM2: See what I mean with my previous comment?

src/shell/pipe_exec/mod.rs
```diff
@@ -431,8 +432,12 @@ impl PipelineExecution for Shell {
         };
 
         // If the given pipeline is a background task, fork the shell.
-        if let Some(command_name) = possible_background_name {
-            fork_pipe(self, piped_commands, command_name)
+        if let Some((command_name, disown)) = possible_background_name {
+            fork_pipe(self, piped_commands, command_name, if disown {
+                ProcessState::Empty
```
jD91mZM2: I have no idea how to disown, I didn't understand the `disown` command's code a lot. So I did this.

mmstick: In this case, you don't need to add the job to the background process table at all. Slots that are marked as `Empty` will be rewritten when a new job is added to the table.

src/parser/pipelines/collector.rs
mmstick: It's fine to keep it as it's own variant. There are ways that we could optimize it this way using bitwise operations.

src/shell/pipe_exec/mod.rs
jD91mZM2: Sorry for being slow. I got busy with other things. I'll look into it tomorrow!

mmstick: Effectively what you'll do is just have the parent do nothing after the fork -- the child will run in the background and the parent will have no knowledge of it.

jD91mZM2: I just ended up conditionally running `send_to_background` in `fork_pipe`. Don't know if that's what you meant, sorry if not.

mmstick:
Thanks!

Issue #552, opened 2017-10-14
A command with a heredoc after a herestring will panic with an array out of bounds variable.


chrisvittal:
**Reproduction**: `cat <<< hello <<EOF`
**Expected behavior**: the ability to enter a heredoc followed by printing something, either "hello" or the heredoc or both.

**Actual behavior**: ion panics and crashes

**Build information**: 
```
$ ion --version
ion 1.0.5 (x86_64-unknown-linux-gnu)
rev e4952a2151d51a273faca8febf63395e32e30874
$ rustc -V
rustc 1.22.0-nightly (a47c9f870 2017-10-11)
```

**Misc**: Backtrace
```
cdv:~/src/ion# cat <<< hello <<EOF
thread 'main' panicked at 'slice index starts at 1 but ends at 0', /checkout/src/libcore/slice/mod.rs:752:4
stack backtrace:
   0: std::sys::imp::backtrace::tracing::imp::unwind_backtrace
             at /checkout/src/libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at /checkout/src/libstd/sys_common/backtrace.rs:69
   2: std::panicking::default_hook::{{closure}}
             at /checkout/src/libstd/sys_common/backtrace.rs:58
             at /checkout/src/libstd/panicking.rs:381
   3: std::panicking::default_hook
             at /checkout/src/libstd/panicking.rs:397
   4: std::panicking::rust_panic_with_hook
             at /checkout/src/libstd/panicking.rs:577
   5: std::panicking::begin_panic
             at /checkout/src/libstd/panicking.rs:538
   6: std::panicking::begin_panic_fmt
             at /checkout/src/libstd/panicking.rs:522
   7: rust_begin_unwind
             at /checkout/src/libstd/panicking.rs:498
   8: core::panicking::panic_fmt
             at /checkout/src/libcore/panicking.rs:71
   9: core::slice::slice_index_order_fail
             at /checkout/src/libcore/slice/mod.rs:752
  10: <core::ops::range::Range<usize> as core::slice::SliceIndex<[T]>>::index
             at /checkout/src/libcore/slice/mod.rs:875
  11: core::slice::<impl core::ops::index::Index<I> for [T]>::index
             at /checkout/src/libcore/slice/mod.rs:728
  12: <alloc::vec::Vec<T> as core::ops::index::Index<core::ops::range::Range<usize>>>::index
[...]

PR #636, opened 2017-12-07
Fix panic on heredoc

AgustinCB:
This fixes the panic on #552.

It's not the final solution, as the logic is still flawed in some cases, but will solve immediate, more urgent problem.

### DIFF
--- a/src/parser/pipelines/collector.rs
+++ b/src/parser/pipelines/collector.rs
@@ -350,11 +350,11 @@ impl<'a> Collector<'a> {
                                 }
                                 unsafe { String::from_utf8_unchecked(buffer) }
                             };
-                            let heredoc = heredoc.lines().collect::<Vec<&str>>();
-                            // Then collect the heredoc from standard input.
-                            inputs.as_mut().map(|x| {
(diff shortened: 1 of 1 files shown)


src/parser/pipelines/collector.rs
```diff
@@ -353,7 +353,7 @@ impl<'a> Collector<'a> {
                             let heredoc = heredoc.lines().collect::<Vec<&str>>();
                             // Then collect the heredoc from standard input.
                             inputs.as_mut().map(|x| {
-                                x.push(Input::HereString(heredoc[1..heredoc.len() - 1].join("\n")))
+                                x.push(Input::HereString(heredoc[1..heredoc.len()].join("\n")))
```
mmstick: Seems that this section should be written as:

```rust
let heredoc = heredoc.lines().skip(1).collect::<Vec<&str>>();
if heredoc.len() > 1 {
    let herestring = Input::HereString(heredoc[..heredoc.len()-1].join("\n"));
    inputs.as_mut().map(|x| {
        x.push(herestring.clone());
    });
}
```

mmstick:
Thanks!

PR #644, opened 2017-12-13
writeln!(stderr, ...) -> eprintln!

AgustinCB:
While working in other tasks, I noticed that there's a rather
inconsistent way to handle errors. In some places we use `eprintln!` and
in others `writeln!` to standard error.

It seems like the use of `writeln!` don't have an strong motive, as the
output is always to standard error, the lock is always handled the same
than `eprintln` and there's no obvious (to me) optimization, so I
decided to use the more idiomatic `eprintln!` everywhere.

Let me know if there's seomthing I'm missing!

### DIFF
--- a/src/builtins/job_control.rs
+++ b/src/builtins/job_control.rs
@@ -5,7 +5,6 @@ use shell::Shell;
 use shell::job_control::{JobControl, ProcessState};
 use shell::signals;
 use shell::status::*;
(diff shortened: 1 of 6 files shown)


mmstick:
Thanks! The reason they were in use was because Ion began development at a time when the `eprintln` macro did not exist, so there's a mixture of `eprintln` and `writeln`.

PR #647, opened 2017-12-13
Add title bar setting, add ability to produce any character using hex code

jackpot51:
This change does the following:

- Adds `${USER}: ${PWD}` to the title bar using the standard xterm escape codes
- Adds a new variable type `hex` or `x` to interpret hex codes as raw characters, allowing any character to be produced as part of an `ion` variable.


mmstick:
Thanks!

PR #648, opened 2017-12-14
Add manpage for ion

Sag0Sag0:


myfreeweb:
That's a help message formatted to look like a manpage :) Check out [help2man](https://www.gnu.org/software/help2man/), it generates a real manpage from a help message.

Sag0Sag0:
True, however the line is blurry. :)

PR #652, opened 2017-12-14
Fix type casts in sys::redox

dlrobertson:
`sys::redox::geteuid` and `sys::redox::getuid` should cast the result of
`sys::redox::cvt`, not the syscall.


jackpot51:
Thanks!

Issue #640, opened 2017-12-08
Quote-Related Parsing Issue


mmstick:
### Input

```
echo "\$test"
```

### Expected

```
$test
```

## Actual

```
\$test
```

covercash2:
this issue is closely related to #641 in that macOS has different behavior. 

initrc:

    fn PROMPT
	    printf "${USER}\@$(hostname)\n > "
    end

in the source, i added a println [here](https://github.com/redox-os/ion/blob/49fba94a6c8738b9c77b6cf059e8c4d6871fdb7d/src/shell/pipe_exec/mod.rs#L751) in shell/pipe_exec/mod.rs fn pipe, i.e.:

`eprintln!("parent: {}", parent.long());`

macOS output:

    parent: hostname
    parent: printf user\\@mac\\ > "  
    user@mac
     > printf "\$hello\n"
    parent: printf \\$hello\\n
    $hello
    parent: hostname
    parent: printf user\\@mac\\ > "
    user@mac
     > echo "\$hello"
    \$hello
    
debian output:

    parent: hostname
    parent: printf user\\@nix\\ > "
    user\@nix
     > printf "\$hello\n"
    parent: printf \\$hello\\n
    \$hello
    parent: hostname
    parent: printf user\\@nix\\ > "
    user\@nix
     > echo "\$hello"
    \$hello

the interesting part is that on macOS printf does the right thing but echo doesn't. and on linux both printf and echo fail to escape correctly.

there's probably a clue in that fact.

PR #657, opened 2017-12-18
Fix issues with escaping on double quote context

gnieto:
**Problem**: 
There was an issue escaping strings on double quotes when a string was being extended.
The issue is described here: https://github.com/redox-os/ion/issues/640

**Solution**: 
Escaping symbols on double-quote environment was not working as
intended. The previous code, when it found a backslash on a quote, it
generated a Word which finished with an escape backslash as text.

With the new code, when a backslash is found a word is emitted up to the
backslash. After this, it checks if the following character should be
escaped: If it shouldn't, we don't advance the cursor and we parse that
backslash as text, on a following word. If it should be escaped, we
advance the buffer, to avoid that the backslash is added to the
following word.

- Changed how words are tokenized when extending a string

**Drawbacks**: 

- The new code, when finds a backslash needs to look ahead the following character to know if needs escaping or not.
- A list of escaped characters is need

**Fixes**: https://github.com/redox-os/ion/issues/640

### DIFF
--- a/src/builtins/test.rs
+++ b/src/builtins/test.rs
@@ -1,9 +1,9 @@
+use super::man_pages::{print_man, MAN_TEST};
 use smallstring::SmallString;
 use std::fs;
 use std::os::unix::fs::{FileTypeExt, MetadataExt, PermissionsExt};
 use std::path::Path;
 use std::time::SystemTime;
-use super::man_pages::{print_man, MAN_TEST};
 
 pub(crate) fn test(args: &[&str]) -> Result<bool, String> {
     let arguments = &args[1..];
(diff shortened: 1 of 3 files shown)


src/builtins/test.rs
```diff
@@ -1,9 +1,9 @@
+use super::man_pages::{print_man, MAN_TEST};
```
gnieto: This change was introduced by rustfmt

mmstick:
This PR seems to cause an issue with the `examples/methods.ion` integration test.

gnieto:
Hi! Yes I saw that the proposed approach has some other issues that I didn't took into account. I'll close this for now, while I think a new way to solve it.

Issue #585, opened 2017-11-12
Panic With `let a = @(())` / `let a = $(( () ))`


xTibor:
**Reproduction**: 
```
let a = @((b))
```

**Actual behavior**:
A panic with this message: `thread 'main' panicked at 'attempt to subtract with overflow', src/parser/arguments.rs:87:25`
[Backtrace](https://gist.github.com/xTibor/635e560c6fc25f9e5c1ba6aa75da18cf)

**Build information**:
`rustc -V`: `rustc 1.23.0-nightly (d6b06c63a 2017-11-09)`
`git rev-parse HEAD`: 2f176ebe1010b4ebd9e2a9031ad354e598cdee9c
System: KDE neon 5.11 (Ubuntu 16.04 LTS)

**Misc**:
The panic is only reproducible in debug mode. In release mode these overflow checks are disabled by the compiler.

xTibor:
A different expression that triggers the same panic:

```
let a = $(( () ))
```

mmstick:
Should be easy to solve. I'm guessing it has something to do with the `))` at the end. A few eprintln! statements should get to the bottom of that. I can't work on this myself at the moment though.

AgustinCB:
Not sure I understand what's the expected output, sorry, could you expand please?

xTibor:
The first example: No idea, most likely a syntax error. As far as I know there's no such expansion as `@(())` in Ion, only `@()`.

The second example: Ion should try to evaluate the `()` expression inside the `$(())` and throw a calc error that it's invalid. (`expected number token, got CloseParen`)

mmstick:
Seems to be resolved.

xTibor:
Checked it again just to make sure it has been fixed, however it's still reproducible for me with 95fefdd8078b618513e84297241606046dcf3357.

mmstick:
@xTibor Which command are you using to get a panic?

xTibor:
The examples I posted above. Both of them still triggers the panic.

mmstick:
This is what happens on my end:

```
mmstick:~/S/ion# ion --version
ion 1.0.5 (x86_64-unknown-linux-gnu)
rev 95fefdd8078b618513e84297241606046dcf3357
mmstick:~/S/ion# echo @((b))
ion: syntax error: '(' at position 1 is out of place

)
mmstick:~/S/ion# echo $(( () ))
expected number token, got CloseParen instead
```

xTibor:
[...]

PR #659, opened 2017-12-22
fix for issue #585

covercash2:
**Solution**: 

the `level: u8` variable in the `Iterator::next` implementation for the `ArgumentSplitter` was being decremented when its value was 0, causing a panic (unsigned ints cannot be less than 0).

**Changes introduced by this pull request**

i assumed that `level` represented the depth of the parentheses, so i consolidated the handling of parentheses and added `level +=1` or `level -= 1` respectively.

i also refactored the raw bit flags to use the bitflags crate for a few reason:
1) the crate was already imported
2) it makes the code a lot more readable (to me at least)
3) it helped me understand the code

**Drawbacks**

i just assumed `level` was referring to the parenthesis level, but i could be wrong. if that's the case this pull request should be rejected.


mmstick:
Thanks!

PR #669, opened 2018-01-04
Call execve from fork_and_exec in sys/redox.rs so they can share code

blackle:
**Problem**: execve and fork_and_exec share a large chunk of code

**Solution**: Remove the argument building and path resolution in fork_and_exec and call execve instead

- There is no more repeated code between fork_and_exec and execve

**Drawbacks**: Ion forks even if the file doesn't exist, whereas before it wouldn't.

**State**: Ready

### DIFF
--- a/src/lib/sys/redox.rs
+++ b/src/lib/sys/redox.rs
@@ -93,91 +93,46 @@ pub(crate) fn fork_and_exec<F: Fn()>(
     clear_env: bool,
     before_exec: F
 ) -> io::Result<u32> {
-    // Construct a valid set of arguments to pass to execve. Ensure
-    // that the program is the first argument.
-    let mut cvt_args: Vec<[usize; 2]> = Vec::new();
-    cvt_args.push([prog.as_ptr() as usize, prog.len()]);
-    for arg in args {
-        cvt_args.push([arg.as_ptr() as usize, arg.len()]);
-    }
-
-    // Get the PathBuf of the program if it exists.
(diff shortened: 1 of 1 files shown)


jackpot51:
This looks good to me, and it is what the standard library does: https://github.com/redox-os/rust/blob/master/src/libstd/sys/redox/process.rs

mmstick:
The separation was done to ensure that the child of the fork that calls **execve** will not allocate any memory, and the issue being that the pre-existing `execve()` function does precisely that when it constructs its arguments. The allocations to construct the arguments need to be done before the fork is called.

jackpot51:
If the allocations are done before, then both the fork and parent will have copies of the allocated memory. Is there a reason it should not be done after?

mmstick:
@jackpot51 I've noticed that the fish shell goes to great lengths to ensure that no allocations occur within the child of a fork. Their rationale being that `malloc` internally uses a mutex when allocating memory for thread safety, yet that mutex can cause issues when you are mixing threads with forks (which Ion is currently doing).

mmstick:
On another note though, I have been thinking about doing away with Ion's usage of threads, so that it doesn't mix threads with forks at all.

jackpot51:
Ok, feel free to revert this if it is not yet a good idea

mmstick:
I'll keep it, but I'll probably separate the argument-generation logic from the argument-execution logic.

PR #672, opened 2018-01-10
Update methods used to print to stdout and stderr

Sag0Sag0:


src/lib/builtins/functions.rs
```diff
@@ -5,15 +5,13 @@ use std::io::{self, Write};
...
-    let stdout = &mut stdout.lock();
-    let _ = writeln!(stdout, "# Functions");
+    println!("# Functions");
     for fn_name in functions.keys() {
         let description = &functions.get(fn_name).unwrap().get_description();
         if let Some(ref description) = *description {
-            let _ = writeln!(stdout, "    {} -- {}", fn_name, description);
+            println!("    {} -- {}", fn_name, description);
         } else {
-            let _ = writeln!(stdout, "    {}", fn_name);
```
mmstick: This is actually done for performance reasons. Each invocation of `println!()` is costly because it locks, writes, and unlocks. Doing it this way only needs to lock once.

src/lib/builtins/mod.rs
```diff
@@ -220,9 +214,7 @@ fn builtin_is(args: &[&str], shell: &mut Shell) -> i32 {
     match is(args, shell) {
         Ok(()) => SUCCESS,
         Err(why) => {
-            let stderr = io::stderr();
-            let mut stderr = stderr.lock();
-            let _ = stderr.write_all(why.as_bytes());
+            eprint!("{}", why);
```
mmstick: These should really be `eprintln!()`.

src/lib/builtins/random.rs
```diff
@@ -17,26 +15,19 @@ fn rand_list(args: &[&str]) -> Result<(), String> {
         output.dedup();
     }
     for out in output {
-        write!(stdout, "{} ", out);
+        print!("{} ", out);
     }
-    writeln!(stdout);
```
mmstick: Same as above -- also done for performance.

src/lib/builtins/functions.rs
Sag0Sag0: jackpot51 suggested doing this. Look at the developer chat. The performance impact is tiny.

src/lib/builtins/mod.rs
Sag0Sag0: FIxing.

mmstick:
I think you forgot to `git add` your changes.

Issue #638, opened 2017-12-08
Array Parsing Issue


mmstick:
### Input
```
echo [1]
echo [123]
```

### Expected

```
1
123
```

### Actual

```


```

PR #673, opened 2018-01-13
Fix to array parsing issue 

spacingnix:
**Solution**:
Glob_check was incorporating one element arrays into globs and somehow consuming them. The fix simply checks if the `b']'` is not succeeded by another character, which would make it a glob pattern.

**Fixes**:
Array Parsing Issue #638


mmstick:
> Unable to locate package x86-64-unknown-redox-gcc

Seems to be an issue with Rust on the Redox test. Can't merge until that's fixed.

jackpot51:
I will see what is happening

mmstick:
Seems to be fixed now. Thanks!

Issue #674, opened 2018-01-14
We should remove script_exec test dependency on examples/a* and examples/b*


aimof:
**Description**
Always when we try to change integration test which file name starts a or b, we must change examples/script_exec.out.

**Expected behavior**
Tests should not depend on each other.

**Actual behavior**
script_exec test depends on a* and b*.

**Blocking/related**
When I write a test to #638, i notice this problem.

**Misc**
I'll make a PR.

PR #675, opened 2018-01-14
Change: remove tests dependency

aimof:
script_exec test is depends on examples/a* and examples/b*.
If we try to change tests in a* or b*, we must change
script_exec.out
So, I remove dependency.

#674 
We must change two test when we add test in examples/a* or examples/b*.

I made directory named examples/script_exec/ copied examples/a* and b*.

We have two solutions to this problem.  The one is mine.  The other is to make examples/script_exec.out on bash from examples/a* and b*.  I tried former because it is not good dependency exists.

- Make examples/script_exec/ and copy a*, b*.
- Change the path in script_exec.ion.

**State**
ready

**Blocking/related**
#638 integration test


mmstick:
Thanks!

PR #676, opened 2018-01-14
CD_CHANGE function

### LINKED ISSUE #400: run command when changing directories
fish functions can trigger when variables (eg `PWD`) are changed; i use this to track my most used directories. shadowing `cd` isn't enough, as that wouldn't trigger on implicit directory changing.

https://fishshell.com/docs/current/commands.html#function

> -v or --on-variable VARIABLE_NAME tells fish to run this function when the variable VARIABLE_NAME changes value.

jD91mZM2:
This would fix #400.
As it happens, this also works with implicit CDs. But I can't intercept every directory change because I don't always have a Shell instance available.


mmstick:
Thanks!

PR #680, opened 2018-01-20
Synchronize Redox and Unix job_control

jackpot51:
This fixes job control on Redox and makes it look more like the Unix version. This also updates the lock file, which is necessary in order to use the wif* functions on Redox.


mmstick:
Thanks!

Issue #684, opened 2018-01-27
When cd'ing into a directory symlink, the path is replaced the absolute path to the symlinked directory


forbjok:
I don't know whether this is intentional or an unwanted side-effect of the way paths are handled, but when changing into a directory symlink, the original symlinked path is lost and replaced with the absolute path to the symlinked directory.

**Expected behavior**:
The symlinked path being preserved, so that doing `cd ..` in the symlinked directory via the symlink will put you back in the directory containing the symlink.
This is what happens in every other shell I've used. (sh, bash, zsh)

The expected behavior as seen in bash:
![screenshot from 2018-01-27 12-46-43](https://user-images.githubusercontent.com/4687674/35471668-3220f2cc-0360-11e8-8f6a-fd7a06e050bb.png)

**Actual behavior**:
if you `cd` into the symlink and then `cd ..`, you end up in the parent directory of the symlinked directory instead of back in the directory containing the symlink.

This screenshot illustrates it:
![screenshot from 2018-01-27 12-30-57](https://user-images.githubusercontent.com/4687674/35471601-acdddfcc-035e-11e8-926d-5214b1072882.png)

**Build information**:
```
$ rustc -V
rustc 1.25.0-nightly (bacb5c58d 2018-01-26)

$ uname -a
Linux ariana 4.14.15-1-ARCH #1 SMP PREEMPT Tue Jan 23 21:49:25 UTC 2018 x86_64 GNU/Linux

$ ion --version
ion 1.0.0-alpha (x86_64-unknown-linux-gnu)
rev 08cf7ed14096459d6ed2efb962e0435159d5928c
```

sourgrasses:
Hmm, isn't this correct behavior? I would expect changing directory to ../ to take me one branch up the tree with reference to the current directory because ../ belongs to the current directory and not to the shell, and in using a symlink you're jumping to another branch on the tree. This is how fish behaves, anyway, and then you can just use the directory history to `cd -` into the previous directory, which I think is an option on bash and works on ion also.

forbjok:
I can't really say whether or not it's objectively correct or not, just that it differs from the behavior of every other shell I know of, and hence from what I would expect. It also seems less practical than how other shells behave, since when you make a symlink you basically want it to behave as if it was just a regular directory in the location of the symlink and this breaks that behavior.

[...]

PR #689, opened 2018-02-01
Fix cd'ing into a symlinked directory losing the original path

forbjok:
Closes https://github.com/redox-os/ion/issues/684

### DIFF
--- a/src/lib/shell/directory_stack.rs
+++ b/src/lib/shell/directory_stack.rs
@@ -235,18 +235,35 @@ impl DirectoryStack {
         dir: &str,
         variables: &Variables,
     ) -> Result<(), Cow<'static, str>> {
-        match (set_current_dir(dir), current_dir()) {
-            (Ok(()), Ok(cur_dir)) => {
-                self.push_dir(cur_dir, variables);
+        use std::path::{Component, Path};
+
+        // Create a clone of the current directory.
+        let mut new_dir = match self.dirs.front() {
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

src/lib/shell/directory_stack.rs
```diff
@@ -235,18 +235,35 @@ impl DirectoryStack {
...
-        match (set_current_dir(dir), current_dir()) {
-            (Ok(()), Ok(cur_dir)) => {
-                self.push_dir(cur_dir, variables);
+        use std::path::{Component, Path};
+
+        // Create a clone of the current directory.
+        let mut new_dir = match self.dirs.front() {
+            Some(cur_dir) => cur_dir.clone(),
+            None => PathBuf::new()
+        };
```
mmstick: Note that you can write this as:

```rust
let new_dir = self.dirs.front().map_or(PathBuf::new(), |dir| dir.clone());
```

forbjok: Won't using map_or cause PathBuf::new() to be evaluated eagerly so that it gets instantiated every time regardless of whether there is a current directory or not?

mmstick: Yeah, `map_or_else` is generally what Clippy will bark at you to use.

mmstick: Which reminds me that Clippy needs to be run on Ion's code base, and all issues reported by Clippy addressed.

PR #694, opened 2018-02-16
Fix color code in setup.ion

PigeonF:
Running `ion setup.ion build docs` failed with an error message about `09F` not being a valid color. The fix is simple, as the cause for this was only a small typo.


mmstick:
Thanks!

Issue #697, opened 2018-02-17
first pushd does not change directory


zen3ger:
**Reproduction**:
```sh
zen3ger:~/Projects/ion# pushd ~
/home/zen3ger /home/zen3ger/Projects/ion
zen3ger:~/Projects/ion# pwd
/home/zen3ger/Projects/ion
zen3ger:~/Projects/ion# pushd
/home/zen3ger/Projects/ion /home/zen3ger
zen3ger:~/Projects/ion# dirs -v
 0  /home/zen3ger/Projects/ion
 1  /home/zen3ger
zen3ger:~/Projects/ion# pwd
/home/zen3ger/Projects/ion
zen3ger:~/Projects/ion# pushd
/home/zen3ger /home/zen3ger/Projects/ion
zen3ger:~# pwd
/home/zen3ger
zen3ger:~# pushd Documents/
Documents/ /home/zen3ger /home/zen3ger/Projects/ion
zen3ger:~# pwd
/home/zen3ger
zen3ger:~# pushd
/home/zen3ger Documents/ /home/zen3ger/Projects/ion
zen3ger:~# pushd
Documents/ /home/zen3ger /home/zen3ger/Projects/ion
zen3ger:~/Documents# pwd
/home/zen3ger/Documents
```

**Expected behavior**:
ion changes directory on first `pushd $NEW_DIR` to `$NEW_DIR`

**Actual behavior**:
After `pushd $NEW_DIR` ion stays at the same directory. When I try to flip the directory order of the 0th and 1st entry with `pushd` it takes two cycles to actually get to the `$NEW_DIR` I've intended to go to.
It seems like after `pushd $NEW_DIR` results in the correct dir stack order of:
0 - NEW_DIR
1 - OLD_DIR
but the `cd $NEW_DIR` never happens when argument is passed to `pushd`.

**Build information**: 
ion 1.0.0-alpha (x86_64-unknown-linux-gnu)
rev 18b6f5eb4a9a10ca404c9fbd302ab5b216c2f45a

**Blocking/related**:
#696

Sag0Sag0:
Fixed.

zen3ger:
Should I close this one?

Sag0Sag0:
Yes.

Issue #696, opened 2018-02-17
pushd on current directory


zen3ger:
**Reproduction**:
```sh
zen3ger:~/Projects/ion# pushd .
. /home/zen3ger/Projects/ion
zen3ger:~/Projects/ion# cd ../
ion: failed to set current dir to : No such file or directory (os error 2)
zen3ger:~/Projects/ion# popd
/home/zen3ger/Projects/ion
zen3ger:~/Projects/ion# cd ../
zen3ger:~/Projects/ion# pwd
/home/zen3ger/Projects
```

**Expected behavior**: 
I should be able to push `.` on the dir stack and back up a directory. After `popd` the prompt should show the current directory path.

**Actual behavior**: 
ion fails to execute `cd ../` on `.` directory. After removing `.` from the dir stack with `popd` the prompt freezez and shows the directory name where `popd` got executed.

**Build information**:
ion 1.0.0-alpha (x86_64-unknown-linux-gnu)
rev 18b6f5eb4a9a10ca404c9fbd302ab5b216c2f45a

Sag0Sag0:
I'm unsure about the correct way to fix this. Should the `pushd` builtin expand `.` and `..` and if not where should they be expanded?

PR #698, opened 2018-02-19
Make pushd change director. Fix issue #697.

Sag0Sag0:


Sag0Sag0:
This test failure doesn't seem to have been caused by me.

src/lib/shell/directory_stack.rs
```diff
@@ -163,6 +163,7 @@ impl DirectoryStack {
             Action::Push(dir) => {
                 let index = if keep_front { 1 } else { 0 };
                 self.insert_dir(index, dir, variables);
```
zen3ger: Could issue #696 be solved if `.` and `..`  get expanded here? In bash that's the default behavior.

Sag0Sag0: It could be however I'm not sure that it should. I feel like `.` and `..` should be expanded long before this point, even before the `pushd` builtin receives them.

mmstick:
I can't merge this until the tests are passing. If you could fix the doc tests that are failing, that'd be great.

mmstick:
Thanks!

PR #700, opened 2018-02-23
Fix doctests

zen3ger:
Add missing `ignore` to doctest of private modules.


zen3ger:
Changes merged in #698

PR #702, opened 2018-03-03
Add dummy --login argument to ion

Sag0Sag0:
**Problem**: Many terminal emulators automatically pass `-l` as an argument. Currently ion does not do anything when it is invoked as a login shell and doesn't check for `-l`. This means that ion tries to execute `-l` as a shell script.

**Solution**: I implemented a dummy `-l`.

### DIFF
--- a/src/lib/shell/binary/mod.rs
+++ b/src/lib/shell/binary/mod.rs
@@ -22,7 +22,7 @@ pub const MAN_ION: &'static str = r#"NAME
     ion - ion shell
 
 SYNOPSIS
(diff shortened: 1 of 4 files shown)


src/lib/shell/binary/mod.rs
```diff
@@ -37,6 +37,9 @@ OPTIONS
 
     -v or --version
         prints the version, platform and revision of ion then exits.
+
+    -l or --login
+        currently does nothing, however in the futere will run ion as a login shell.
```
mmstick: There's a typo here.

Issue #708, opened 2018-03-11
Asterisk tilde (*~) acts as asterisk space tilde (* ~)


kirch7:
**Reproduction**: Suppose you are in a directory `$HOME/IC/superboids/`. In this directory there are some files endding in tilde (e.g. `pineapple.sh~`). Now you want to remove theses files using `rm`.

**Expected behavior**:
`rm *~` should remove all files in the current directory whose names end in tilde.
```
kirch:~/IC/superboids# ls
b	   lines.py   Makefile	     onion.py	   pineapple.sh~  video2types_.gp  video.gp	 video.py	virtuais3.avi
lines.dat  lines.py~  Miniboid.cpp~  pineapple.sh  progress.txt   video2types.gp   video_inf.gp  virtuais2.avi	virtuais.avi
```
```
kirch:~/IC/superboids# rm *~
```
```
kirch:~/IC/superboids# ls
b	   lines.py   Makefile	     onion.py	   video2types_.gp  video.gp	 video.py	virtuais3.avi
lines.dat  pineapple.sh  progress.txt   video2types.gp   video_inf.gp  virtuais2.avi	virtuais.avi
```

**Actual behavior**:
`rm *~` actually acts as `rm * ~`.
```
kirch:~/IC/superboids# ls
b	   lines.py   Makefile	     onion.py	   pineapple.sh~  video2types_.gp  video.gp	 video.py	virtuais3.avi
lines.dat  lines.py~  Miniboid.cpp~  pineapple.sh  progress.txt   video2types.gp   video_inf.gp  virtuais2.avi	virtuais.avi
```
```
kirch:~/IC/superboids# rm *~
rm: cannot remove '.git': Is a directory
rm: cannot remove 'b': Is a directory
rm: cannot remove '/home/kirch': Is a directory
```
```
kirch:~/IC/superboids# ls
b
```

**Build information**:
ion 1.0.0-alpha (x86_64-unknown-linux-gnu)
rev 96d5b5d2ef53d8d9e93a4a41019560175cc12598

**Related**:
https://github.com/redox-os/ion/issues/406
https://github.com/redox-os/ion/pull/407

**Misc**: I had similar problems with `ls`. I had backup, there was no huge problem.

ishanjain28:
This is a bug related to #709 and it has been fixed in PR #710

Issue #709, opened 2018-03-12
tilde ~ is expanded multiple times, even if it is not part of a path


zwieberl:
**Reproduction**: 
Just type `touch ~/bla~`

**Expected behavior**:
Create a new file called `bla~` in /home

**Actual behavior**:
```
# touch ~/bla~
touch: cannot touch ‘/home/zwieberl/bla/home/zwieberl’: No such file or directory
```

**Build information**:
rustc -V:
rustc 1.25.0-nightly (3ec5a99aa 2018-02-14)

./target/release/ion --version:
ion 1.0.0-alpha (x86_64-unknown-linux-gnu)
rev 96d5b5d2ef53d8d9e93a4a41019560175cc12598

**Blocking/related**:
#708

PR #710, opened 2018-03-13
Fixed bug in tilde resolution

ishanjain28:
**Problem**:
The problem exists in tilde resolution. Only the first tilde should be expanded, But the current code expands all tildes. 

**Solution**:
My solution checks current position and only sets tilde, If current position is 0. i.e. `~` is the first character.

- Only the first tilde is resolved instead of all tildes in a string.

**Fixes**:
#709 #708 

Completed.

------

### DIFF
--- a/src/lib/parser/shell_expand/words/mod.rs
+++ b/src/lib/parser/shell_expand/words/mod.rs
@@ -636,6 +636,7 @@ impl<'a, E: Expander + 'a> Iterator for WordIterator<'a, E> {
         let mut start = self.read;
         let mut glob = false;
         let mut tilde = false;
+
         loop {
             if let Some(character) = iterator.next() {
                 match character {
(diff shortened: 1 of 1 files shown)


mmstick:
This causes the `parser::shell_expand::words::tests::test_words` test to fail, so it can't be merged yet.

ishanjain28:
@mmstick Sorry, I broke `WordToken` to fix this bug which resulted in one test failing testcase. 

**Problem**:
`~` was used to seperate a string into multiple `WordToken`, So, `~/usr~` would resolve to `[WordToken::Normal("~/usr", false, true), WordToken::Normal("~", false, true)` which would expand to `/home/ishan/usr/home/ishan`. 

**Fix**:
Comment out code that uses `~` as a seperator. 

I also added some comments that explain the reason behind commenting that code. 

Should I completely remove the code?

mmstick:
You should complete remove the code. Commented code just increases clutter.

ishanjain28:
Okay, I'll remove the code.

mmstick:
Thanks!

ishanjain28:
Please close #708  and #709 as both issues have been fixed by this patch. 

Glad to be useful, Looking forward to contributing more. :)

PR #713, opened 2018-03-16
Add builtin type

Sag0Sag0:


src/lib/builtins/command_info.rs
```diff
@@ -0,0 +1,85 @@
...
+use std::env;
+use std::path::Path;
+
+pub(crate) fn which(args: &[&str], shell: &mut Shell) -> Result<i32, String> {
+    if check_help(args, MAN_WHICH) {
+        return Ok(SUCCESS)
+    }
+
+    if args.len() == 1 {
+        return Err(String::from("which: Expected at least 1 args, got only 0"))
```
mmstick: May want to just write this directly to standard error, and change the error return value to `()` to avoid the allocation.

src/lib/builtins/command_info.rs
```diff
@@ -0,0 +1,85 @@
...
+            result = FAILURE;
+        }
+    }
+    Ok(result)
+}
+
+pub(crate) fn find_type(args: &[&str], shell: &mut Shell) -> Result<i32, String> {
+    // Type does not accept help flags, aka "--help".
+    if args.len() == 1 {
+        return Err(String::from("which: Expected at least 1 args, got only 0"))
```
mmstick: same as above

src/lib/builtins/command_info.rs
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    let mut result = FAILURE;
+    for &command in &args[1..] {
+        let c_type = get_command_info(command, shell);
+        if c_type.is_ok() {
+            match c_type.unwrap().as_str() {
```
mmstick: also the same as above

src/lib/builtins/command_info.rs
```diff
@@ -0,0 +1,85 @@
...
+    Ok(result)
+}
+
+pub(crate) fn get_command_info(command: &str, shell: &mut Shell) -> Result<String, ()> {
+    if shell.variables.aliases.get(command).is_some() {
+        return Ok(String::from("alias"))
+    } else if shell.functions.contains_key(command) {
+        return Ok(String::from("function"))
+    } else if shell.builtins.contains_key(command) {
+        return Ok(String::from("builtin"))
```
mmstick: May want to make the Ok return value a ` Cow<'a, str>` type so that these don't require an allocation.

src/lib/builtins/command_info.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    if args.len() == 1 {
+        return Err(String::from("which: Expected at least 1 args, got only 0"))
+    }
+
+    let mut result = SUCCESS;
+    for &command in &args[1..] {
+        let c_type = get_command_info(command, shell);
+        if c_type.is_ok() {
+            match c_type.unwrap().as_str() {
```
mmstick: Why not just write this?
```rust
if let Ok(c_type) = get_command_info(command, shell) {
    match c_type.as_str() {
        ...
    }
}

File: src/lib/builtins/command_info.rs (hunk 4 above)
Comment (Sag0Sag0) [reply src/lib/builtins/command_info.rs]: Could you give me some pointers on how to do that? I keep on having ownership problems turning
```
let executable = Path::new(path).join(command);
            if executable.is_file() {
                return Ok(executable.display().to_string())
            }
```
into a `Cow<'a, str>`.

mmstick: @Sag0Sag0 https://play.rust-lang.org/?gist=5bf670b09221399daa0ab46d10684ac1&version=stable

Sag0Sag0: @mmstick Thanks for pointing me to that, it made things a lot clearer.

mmstick:
Thanks!

PR #714, opened 2018-03-22
Alias reimplementation  #711

ishanjain28:
Alias Reimplementation as discussed in #711 

- Removed Math Expressions parser from `parse_alias`. 
- Alias reimplemented as a keyword exactly like `let` or `export`. 

**Drawbacks**: 
I have made an assumption in my code that might be considered a drawback.

The assumption is that, Any value stored in `alias` will always resolve to a `Statement::Pipeline`, I don't know what should be the appropriate behaviour if it does not resolves to a `Statement::Pipeline`, So, It silently ignores that `PipeItem`. 

**TODOs**:
- Tests(I'll add those if and once this PR is approved)

**Fixes**:
#711 

**State**:
WIP, Everything except for the tests is done.

### DIFF
--- a/src/lib/builtins/mod.rs
+++ b/src/lib/builtins/mod.rs
@@ -27,7 +27,7 @@ use self::man_pages::*;
 use self::source::source;
 use self::status::status;
 use self::test::test;
-use self::variables::{alias, drop_alias, drop_array, drop_variable};
+use self::variables::{drop_alias, drop_array, drop_variable};
 use types::Array;
 
 use std::env;
@@ -71,7 +71,6 @@ macro_rules! map {
 
 /// Builtins are in A-Z order.
 pub const BUILTINS: &'static BuiltinMap = &map!(
-    "alias" => builtin_alias : "View, set or unset aliases",
     "and" => builtin_and : "Execute the command if the shell's previous status is success",
     "bg" => builtin_bg : "Resumes a stopped background process",
     "bool" => builtin_bool : "If the value is '1' or 'true', return 0 exit status",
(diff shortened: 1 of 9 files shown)


ishanjain28:
Apologies for some poor commit messages, I expect the commits to be squashed together.

mmstick:
Any chance you could look into expanding upon it so that it can store a `Vec<Statement>`? It should be possible to do this, for example:

```
$ alias testing = 'echo one; for x in [two, three]; echo $x; end'
$ testing
one
two
three
$ _
```

ishanjain28:
I'll just cover working of my code, 
    alias ll=/usr/bin/ls -l 

`ll` will resolve to `Pipeline{PipeItem{..., command: "ll", Kind: Last}}`, When it goes to `execute_toplevel`, The following 5 changes are made to each element of pipeline.

Check if the command is an alias, If it is, Do the following,
1. Call `StatementSplitter` on the value stored in alias key. 
2. Remove the original PipeItem(One which had the alias as a command, Here on referred to as "Original Pipeitem")
3. Put the new Pipeitems in it's place that were created by calling `StatementSplitter` on alias value.
4. Take args from Original Pipeitem and append them to the last newly created Pipeitem.
5. Change the kind of last newly created PipeItem to that of Original PipeItem.

ishanjain28:
@mmstick I'll try to add support for that. I came up with a different version of this code yesterday that involved calling `execute_toplevel` on each PipeItem generated after expanding alias instead of replacing the original pipeitem but that didn't work well. 

Any pointers on how can I implement that cleanly?

mmstick:
My knowledge is a bit rusty. Let me look through the parsing code to see how I've been doing it.

ishanjain28:
Okay, Cool.

mmstick:
Something like this should get you what you want.

```rust
let mut statements = Vec::new();
for result in StatementSplitter::new(alias_value).map(Statement::parse) {
    statements.push(result?);
}
```

PR #716, opened 2018-04-07
fix builtin crash for and,or,not

jcaplan:

running `not`, `or`, `and` with no arguments results in a crash.

keep it simple. `not <nothing> = FAILURE`

error message is printed e.g.
```
~/redox/ion/target/release# and
and requires at least one argument
```

### DIFF
--- a/src/lib/builtins/mod.rs
+++ b/src/lib/builtins/mod.rs
@@ -559,6 +559,10 @@ fn builtin_not(args: &[&str], shell: &mut Shell) -> i32 {
     if check_help(args, MAN_NOT) {
         return SUCCESS;
     }
+    if args.len() < 2 {
+        eprintln!("not requires at least one argument");
+        return FAILURE;
+    }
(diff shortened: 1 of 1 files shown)


src/lib/builtins/mod.rs
```diff
@@ -559,6 +559,12 @@ fn builtin_not(args: &[&str], shell: &mut Shell) -> i32 {
     if check_help(args, MAN_NOT) {
         return SUCCESS;
     }
+    if args.len() < 2 {
```
panaman67: this logic seems like it could be made into a method? As its more or less the same code except for the name of the arg. I could be wrong though. Thoughts?

jcaplan: it does seem a bit complicated. I copied the template from `builtin_bool`. I'm not really sure what the need is for the locking. I'm now noticing most of the other functions just use `eprintln`. I'll simplify the block but for such short functions I lean aesthetically towards leaving it inline.

jcaplan:
abandoning in favour of #718

PR #718, opened 2018-04-08
and/or followed by let results in command not found (fixes #619)

### LINKED ISSUE #619: and/or followed by let results in command not found
**Reproduction**:

```
echo "This does not have a bad exit status"; and let x = "something"
```

**Expected behavior**:
`x` is set to `something`

**Actual behavior**:
```
ion: command not found: let
```

**Build information**:

A `git bisect` showed that this started at 7d9584ee254b6019343d034a4de07e2b5c3e19bd

jcaplan:
and/or must be converted into keywords
followed suggestions in the issue discussion. Copied `Statement::Time` template

### DIFF
--- a/src/lib/parser/statement/parse.rs
+++ b/src/lib/parser/statement/parse.rs
@@ -219,6 +219,14 @@ pub(crate) fn parse(code: &str) -> Statement {
             return Statement::Time(Box::new(parse(cmd[4..].trim_left())))
         }
         _ if cmd.eq("time") => return Statement::Time(Box::new(Statement::Default)),
+        _ if cmd.starts_with("and ") => {
+            return Statement::And(Box::new(parse(cmd[3..].trim_left())))
+        }
+        _ if cmd.eq("and") => return Statement::And(Box::new(Statement::Default)),
+        _ if cmd.starts_with("or ") => {
+            return Statement::Or(Box::new(parse(cmd[2..].trim_left())))
+        }
+        _ if cmd.eq("or") => return Statement::Or(Box::new(Statement::Default)),
         _ => (),
     }
 
--- a/src/lib/parser/statement/parse.rs
+++ b/src/lib/parser/statement/parse.rs
@@ -227,6 +227,10 @@ pub(crate) fn parse(code: &str) -> Statement {
             return Statement::Or(Box::new(parse(cmd[2..].trim_left())))
         }
         _ if cmd.eq("or") => return Statement::Or(Box::new(Statement::Default)),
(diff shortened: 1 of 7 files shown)


mmstick:
Thanks for doing this. Have been wanting this done for quite some time. You could also throw Not in there too. It's pretty similar.

jcaplan:
I seem to have run into another syntax problem:
```
jonah:~/redox/ion/target/debug# true && not false
ion: command not found: not
jonah:~/redox/ion/target/debug# true && and echo hi
ion: command not found: and
jonah:~/redox/ion/target/debug# true && let x=3
ion: command not found: let
jonah:~/redox/ion/target/debug# true && time
jonah:~/redox/ion/target/debug# time
real    0.000001572s
```
for some reason `time` is not throwing the same error but it is also not running... I might have to actually dive into some of that syntax definition.

mmstick:
This is to be expected, for the moment. My idea has been that we need to also treat `&&` and `||` the same as we treat the `and` and `or` keywords. Basically, the fix for this is to modify the `StatementSplitter` to split valid instances of `&&` and `||` into `And` and `Or` statements.

src/lib/parser/statement/parse.rs
```diff
@@ -219,6 +219,18 @@ pub(crate) fn parse(code: &str) -> Statement {
             return Statement::Time(Box::new(parse(cmd[4..].trim_left())))
         }
         _ if cmd.eq("time") => return Statement::Time(Box::new(Statement::Default)),
+        _ if cmd.starts_with("and ") => {
+            return Statement::And(Box::new(parse(cmd[3..].trim_left())))
+        }
```
mmstick: I wonder if we could make a macro to simplify how this is currently done here

mmstick: Basically, rather than forcing an allocation here to make it an owned string, you could alter the return type so that it returns `Statement`s instead of `String`s, which will require some changes to fix tests and such.

jcaplan: After having a look I don't think it changes much and still doesn't solve the main problem I was having in the first place with `data: &str` was figuring out how to replace the data field with the modified statement string and I couldn't seem to navigate my way through all the compiler errors...

mmstick: Basically, there's no need to modify the string. You can go ahead with converting it directly into a Statement, rather than returning an owned string to later be converted into a Statement.

mmstick: I'll likely go ahead and merge this and fix it soon.

PR #720, opened 2018-04-16
Increase default HISTFILE_SIZE to 100000

theduke:
I recently reached the default HISTFILE_SIZE limit of 1000 and actually had to spend a bit of time figuring out what the problem is.

I think the current default of 1000 is much too low.
There's really no reason to constrain this to a low value as far as I can see.

I know other shells don't even disable history by default or also have very low limits, but those have historical bagge of running on very small systems.

100_000 seems much more reasonable to me.


mmstick:
As soon as build issues are resolved, I'll merge this (I can't merge until Travis checks out). May just disable Redox in Travis for now.

mmstick:
Thanks!

PR #725, opened 2018-05-16
601 filename completion

### LINKED ISSUE #601: zsh-like autocompletion (paths)
On both ion and zsh, I can write something like ````cd rust-i<tab>```` and it will autocomplete to ````cd rust-ion````, however zsh also supports path autocompletion between ````/````s, let's say I have /home/matthias/vcs/github/rust, I can write cd ````/h/m/v/g/r<tab>```` and it will autocomplete to ````/home/matthias/vcs/github/rust```` if that is the only match found.
Even /////rust would work if I only had a single "rust" directory at that directory depth in any subfolder (usually this is a silly idea though since zsh would query the FS for all the folder names which causes longer IO wait than actually typing a few letters of the path manually).

Anyway this saves me quite some typing, would be nice to have that in ion as well, in my opinion. :)

emturner:
**Problem**:
Filename completion is only implemented on the final directory/file in the path - it would save time (and typing) to have the completion operate on all the directories in the path concurrently.

**Solution**: 
An extra completion step, using glob::glob is inserted before being given to Liner::Completion (necessary to preserve prior completion behaviour)

- The path is reformatted to insert an asterisk before every '/' [except where the path begins with '/'] and run through the glob completion

**Drawbacks**:
There are now two completion steps, so there may occassionally be a slight slowdown in completion time.  This doesn't seem to be noticeable, unless a path such as '////x/y' is given.

**TODO**:
When using the '...//...' pattern, hidden directories are ignored.  It could be possible to make this an option?

**Fixes**:
fixes #601

**State**:
Ready


mmstick:
Odd that I didn't notice this PR. I will attempt to merge this today. Thanks!

PR #733, opened 2018-05-25
Add support for padded range expansion.

ProgVal:
**Problem**: I wanted `{03..07}` to expand to `03 04 05 06`, as on bash or zsh.

**Solution**: I added a new argument `nb_digits` to range-generating functions, so the output number is padded to get the expected number of digits. Note that `-` counts as a digits in order to get proper alignment when dealing with both negative and positive numbers (like bash and zsh do).

- The behavior of ranges like `{03..07}` changes.

**Drawbacks**: slightly slower, because we have to do some lookup (of at most two characters) and `format!` uses a loop to pad

**TODOs**: nothing

**Fixes**: my personal workflow when I want to match files numbered with leading zeroes to sort alphabetically

**State**: ready

**Blocking/related**: none

**Other**: none

### DIFF
--- a/src/lib/parser/shell_expand/ranges.rs
+++ b/src/lib/parser/shell_expand/ranges.rs
@@ -1,6 +1,6 @@
 use super::words::{Index, Range};
 
-fn stepped_range_numeric(mut start: isize, end: isize, step: isize) -> Option<Vec<String>> {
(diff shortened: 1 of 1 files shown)


mmstick:
Thanks!

PR #739, opened 2018-05-27
Shell Expand Ranges Refactor

stratact:


src/lib/parser/shell_expand/ranges.rs
```diff
@@ -1,57 +1,67 @@
...
-            start += step;
+            Some(Box::new((end..start).scan(start, move |index, _| {
+                if *index > end {
+                    let index_holder = *index;
+                    *index += step; // This step subtracts
+                    Some(format!("{:0width$}", index_holder, width=nb_digits))
+                } else {
+                    None
+                }
+            })))
```
mmstick: This looks like it'd be better written as:

```rust
let (start, end, ordering) = if start < end {
    (start, end, Ordering::Greater)
} else {
    (end, start, Ordering::Less)
};

let iter = (start..end).scan(start, move |index, _| {
    |a: isize, b: isize| -> bool { a < b }
        if end.cmp(index) == ordering {
            let index_holder = *index;
            *index += step; // This step adds
            Some(format!("{:0width$}", index_holder, width=nb_digits))
        } else {
            None
        }
    }
);

Some(Box::new(iter))
```

src/lib/parser/shell_expand/ranges.rs
```diff
@@ -1,57 +1,67 @@
...
-            start = step_func(start, step);
+            Some(Box::new((end..start).scan(start, move |index, _| {
+                if *index > end {
+                    let index_holder = *index;
+                    *index = index.wrapping_sub(step);
+                    Some((index_holder as char).to_string())
+                } else {
+                    None
+                }
+            })))
```
mmstick: Similar means could be used here.

src/lib/parser/shell_expand/ranges.rs
stratact: Doing this fails the test though
```rust
failures:

---- parser::shell_expand::ranges::range_expand stdout ----
thread 'parser::shell_expand::ranges::range_expand' panicked at 'assertion failed: `(left == right)`
  left: `[]`,
 right: `["3", "2", "1", "0", "-1", "-2", "-3"]`', src/lib/parser/shell_expand/ranges.rs:416:5
note: Run with `RUST_BACKTRACE=1` for a backtrace.


failures:
    parser::shell_expand::ranges::range_expand

test result: FAILED. 192 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out
```

stratact: This is probably due to shadowing the `end` variable when I shouldn't.

PR #746, opened 2018-05-29
Use iterators for color attributes

stratact:
Note: Because `Option<Box<Iterator<Item = &'a str> + 'a>>` cannot derive both `Debug` and `PartialEq` some of the tests had to have been commented out, unfortunately.


stratact:
I can understand if you don't want to merge this.

PR #749, opened 2018-05-30
Yet another refactor

stratact:
For the builtin functions, Ion was converting existing `&[String]` values into `Vec<&str>` by collecting each item and converting them with `as_ref()` which makes them `&str` values, to avoid from more collecting, just pass the `&[String]` borrows to save from more allocation and when a function needs to index a `&[String]` just `Deref` and borrow it.

### DIFF
--- a/src/lib/shell/job.rs
+++ b/src/lib/shell/job.rs
@@ -192,12 +192,6 @@ macro_rules! set_field {
     };
 }
 
-fn collect_args(args: &[String]) -> SmallVec<[&str; 16]> {
-    args.iter()
-        .map(|x| x as &str)
-        .collect::<SmallVec<[&str; 16]>>()
-}
-
 impl RefinedJob {
     /// Returns a long description of this job: the commands and arguments
     pub(crate) fn long(&self) -> String {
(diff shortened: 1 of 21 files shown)


src/lib/shell/job.rs
```diff
@@ -242,8 +236,7 @@ impl RefinedJob {
                 ref stdout,
                 ref stderr,
             } => {
-                let args = collect_args(&args);
```
stratact: All for the sake of here

src/lib/shell/mod.rs
```diff
@@ -281,8 +281,7 @@ impl<'a> Shell {
                     Some(SUCCESS)
                 } else {
                     let borrowed = &pipeline.items[0].job.args;
-                    let small: SmallVec<[&str; 4]> = borrowed.iter().map(|x| x as &str).collect();
```
stratact: And here

src/lib/shell/pipe_exec/mod.rs
```diff
@@ -975,7 +975,6 @@ fn spawn_proc(
             ref stderr,
             ref stdin,
         } => {
-            let args: Vec<&str> = args.iter().map(|x| x as &str).collect();
```
stratact: And here

stratact:
Can I get a :hand: ?

mmstick:
Thanks!

PR #751, opened 2018-05-30
Improve StatementSplitter

stratact:
This is an attempt to have `StatementSplitter` migrate back to returning `&str`s but this needs to be reviewed because this change may make ion either make you breakfast or eat your cat :stuck_out_tongue_winking_eye:

### DIFF
--- a/src/lib/parser/statement/mod.rs
+++ b/src/lib/parser/statement/mod.rs
@@ -12,8 +12,8 @@ use shell::flow_control::Statement;
 /// `Statement`
 pub(crate) fn parse_and_validate<'a>(statement: Result<StatementVariant, StatementError>) -> Statement {
     match statement {
-        Ok(StatementVariant::And(statement)) => parse(statement),
-        Ok(StatementVariant::Or(statement)) => parse(statement),
+        Ok(StatementVariant::And(statement)) => Statement::And(Box::new(parse(statement))),
+        Ok(StatementVariant::Or(statement)) => Statement::Or(Box::new(parse(statement))),
         Ok(StatementVariant::Default(statement)) => parse(statement),
         Err(err) => {
             eprintln!("ion: {}", err);
--- a/src/lib/parser/statement/mod.rs
+++ b/src/lib/parser/statement/mod.rs
@@ -4,15 +4,17 @@ mod parse;
 mod splitter;
 
 pub(crate) use self::{
-    parse::parse, splitter::{StatementError, StatementSplitter},
+    parse::parse, splitter::{StatementVariant, StatementError, StatementSplitter},
 };
 use shell::flow_control::Statement;
 
 /// Parses a given statement string and return's the corresponding mapped
 /// `Statement`
-pub(crate) fn parse_and_validate<'a>(statement: Result<String, StatementError>) -> Statement {
+pub(crate) fn parse_and_validate<'a>(statement: Result<StatementVariant, StatementError>) -> Statement {
     match statement {
-        Ok(statement) => parse(statement.as_str()),
+        Ok(StatementVariant::And(statement)) => parse(statement),
+        Ok(StatementVariant::Or(statement)) => parse(statement),
+        Ok(StatementVariant::Default(statement)) => parse(statement),
         Err(err) => {
             eprintln!("ion: {}", err);
             Statement::Error(-1)
--- a/src/lib/parser/statement/splitter.rs
+++ b/src/lib/parser/statement/splitter.rs
@@ -122,18 +122,29 @@ impl<'a> StatementSplitter<'a> {
     }
 
     fn get_statement(&mut self, new_flag: Flags) -> StatementVariant<'a> {
-        let variant = if self.flags.contains(Flags::AND) {
-                self.flags = self.flags - Flags::AND | new_flag;
-                StatementVariant::And(&self.data[self.start + 1..self.read - 1].trim())
-            } else if self.flags.contains(Flags::OR) {
-                self.flags = self.flags - Flags::OR | new_flag;
-                StatementVariant::Or(&self.data[self.start + 1..self.read - 1].trim())
-            } else {
-                self.flags |= new_flag;
-                let statement = &self.data[self.start..self.read - 1].trim();
-                StatementVariant::Default(statement)
-            };
-        variant
+        if self.flags.contains(Flags::AND) {
+            self.flags = (self.flags - Flags::AND) | new_flag;
+            StatementVariant::And(&self.data[self.start + 1..self.read - 1].trim())
+        } else if self.flags.contains(Flags::OR) {
+            self.flags = (self.flags - Flags::OR) | new_flag;
+            StatementVariant::Or(&self.data[self.start + 1..self.read - 1].trim())
+        } else {
+            self.flags |= new_flag;
+            let statement = &self.data[self.start..self.read - 1].trim();
+            StatementVariant::Default(statement)
+        }
+    }
+
+    fn get_statement_from(&mut self, input: &'a str) -> StatementVariant<'a> {
+        if self.flags.contains(Flags::AND) {
+            self.flags -= Flags::AND;
+            StatementVariant::And(input)
+        } else if self.flags.contains(Flags::OR) {
+            self.flags -= Flags::OR;
+            StatementVariant::Or(input)
+        } else {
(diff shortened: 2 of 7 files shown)


src/lib/parser/statement/splitter.rs
```diff
@@ -109,18 +120,33 @@ impl<'a> StatementSplitter {
...
-    fn next(&mut self) -> Option<Result<String, StatementError>> {
-        let start = self.read;
+    fn next(&mut self) -> Option<Self::Item> {
+        self.start = self.read;
         let mut first_arg_found = false;
         let mut else_found = false;
         let mut else_pos = 0;
         let mut error = None;
-        let c = self.data.clone();
+        let c = self.data;
```
mmstick: This variable can be removed.

src/lib/parser/statement/mod.rs
```diff
@@ -4,15 +4,17 @@ mod parse;
...
 use shell::flow_control::Statement;
 
 /// Parses a given statement string and return's the corresponding mapped
 /// `Statement`
-pub(crate) fn parse_and_validate<'a>(statement: Result<String, StatementError>) -> Statement {
+pub(crate) fn parse_and_validate<'a>(statement: Result<StatementVariant, StatementError>) -> Statement {
     match statement {
-        Ok(statement) => parse(statement.as_str()),
+        Ok(StatementVariant::And(statement)) => parse(statement),
+        Ok(StatementVariant::Or(statement)) => parse(statement),
```
mmstick: You'll need to create a Statement::And/Or directly, rather than passing it off to be parsed into a Statement.

src/lib/parser/statement/splitter.rs
```diff
@@ -109,18 +120,33 @@ impl<'a> StatementSplitter {
             math_paren_level: 0,
         }
     }
+
+    fn get_statement(&mut self, flags: Flags) -> StatementVariant<'a> {
+        self.flags.insert(flags);
```
mmstick: The active flag should be set after getting the variant.

src/lib/parser/statement/splitter.rs
```diff
@@ -109,18 +120,33 @@ impl<'a> StatementSplitter {
...
+        match self.flags {
+            Flags::AND => {
+                StatementVariant::And(&self.data[..self.read - 1].trim())
+            }
+            Flags::OR => {
+                StatementVariant::Or(&self.data[..self.read - 1].trim())
+            }
```
mmstick: Don't forget to unset the AND / OR flag after they've been detected.

src/lib/parser/statement/splitter.rs
```diff
@@ -299,25 +314,25 @@ impl Iterator for StatementSplitter {
...
+                        None => Some(Ok(statement)),
                     };
                 }
                 b' ' if else_found => {
                     let output = &self.data[else_pos..self.read - 1].trim();
                     if !output.is_empty() {
                         if "if" != *output {
                             self.read = else_pos;
-                            return Some(Ok(String::from("else")));
+                            return Some(Ok(StatementVariant::Default("else")));
```
mmstick: Don't forget to reset the and/or flags to off here.

src/lib/parser/statement/splitter.rs
```diff
@@ -109,18 +120,33 @@ impl<'a> StatementSplitter {
             math_paren_level: 0,
         }
     }
+
+    fn get_statement(&mut self, flags: Flags) -> StatementVariant<'a> {
+        self.flags.insert(flags);
+        match self.flags {
```
mmstick: This also won't work because flags contains many flags. You'll need to switch to `if flags.contains(Flags::AND) / else if flags.contains(Flags::OR) / else`

src/lib/parser/statement/mod.rs
stratact: How do I do that, since `Statement::{And,Or}` both take in a `Box<Statement>`?

src/lib/parser/statement/splitter.rs
stratact: Will do

src/lib/parser/statement/mod.rs
mmstick:
```rust
Box::new(Statement::And(parse(statement)))
```

stratact: Thank you

stratact:
Err, I think the examples test is stuck due to how recursive `And` and `Or` is...

src/lib/parser/statement/splitter.rs
```diff
@@ -109,19 +120,40 @@ impl<'a> StatementSplitter {
             math_paren_level: 0,
         }
     }
+
+    fn get_statement(&mut self, character: u8) -> StatementVariant<'a> {
+        let variant = if self.flags.contains(Flags::AND) {
+                self.flags.remove(Flags::AND);
+                self.data = &self.data[self.read + 1..].trim();
+                StatementVariant::And(&self.data)
```
mmstick: This doesn't look like it's going to split statements correctly. It's going to return everything after `self.read + 1`, rather than up to.

src/lib/parser/statement/splitter.rs
```diff
@@ -109,19 +120,40 @@ impl<'a> StatementSplitter {
             math_paren_level: 0,
         }
     }
+
+    fn get_statement(&mut self, character: u8) -> StatementVariant<'a> {
```
mmstick: Should use `Flags` instead of `u8`.

src/lib/parser/statement/splitter.rs
```diff
@@ -109,19 +120,40 @@ impl<'a> StatementSplitter {
...
+        let variant = if self.flags.contains(Flags::AND) {
+                self.flags.remove(Flags::AND);
```
mmstick: self.flags = self.flags - Flags::AND + new_flag;

src/lib/parser/statement/splitter.rs
```diff
@@ -109,19 +120,40 @@ impl<'a> StatementSplitter {
...
+                self.flags.remove(Flags::OR);
+                self.data = &self.data[self.read + 1..].trim();
+                StatementVariant::Or(&self.data)
+            } else {
+                if character == b'&' {
+                    self.flags.insert(Flags::AND);
+                }
+                if character == b'|' {
+                    self.flags.insert(Flags::OR);
+                }
```
mmstick: These two if conditions can be removed, and replaced with

```rust
self.flags |= new_flag;
```

src/lib/parser/statement/splitter.rs
stratact: At that point `self.read` just points to `& echo true` in `matches Foo '([A-Z])\w+' && echo true`, so it makes sense just to returns everything after that which is `" echo true".trim()`

src/lib/parser/statement/splitter.rs
```diff
@@ -299,25 +304,27 @@ impl Iterator for StatementSplitter {
...
                     };
                 }
                 b' ' if else_found => {
                     let output = &self.data[else_pos..self.read - 1].trim();
                     if !output.is_empty() {
                         if "if" != *output {
                             self.read = else_pos;
-                            return Some(Ok(String::from("else")));
+                            self.flags.remove(Flags::AND);
+                            self.flags.remove(Flags::OR);
```
mmstick: It's better to write this as

```rust
self.flags.remove(Flags::AND | Flags::OR);
```

src/lib/parser/statement/splitter.rs
stratact: will do

src/lib/parser/statement/splitter.rs
stratact: will do

src/lib/parser/statement/splitter.rs
stratact: will do

src/lib/parser/statement/splitter.rs
stratact: will do

src/lib/parser/statement/splitter.rs
mmstick: Shells can have more elaborate statement-splitting requirements, such as

```sh
cmd arg; cmd arg && cmd this || cmd that; cmd arg
```

PR #754, opened 2018-06-01
Adding `prompt` builtin for selecting or customizing the Ion prompt

stratact:
This PR is **work-in-progress** and of need in guidance to get things working together as a builtin.

### DIFF
--- a/src/lib/builtins/prompt.rs
+++ b/src/lib/builtins/prompt.rs
@@ -1,4 +1,4 @@
-use shell::Shell;
+use shell::{colors::COLORS, Shell};
 use std::io::{self, Write, StdoutLock};
 
 trait AskColor {
@@ -7,11 +7,15 @@ trait AskColor {
 
 impl<'a> AskColor for StdoutLock<'a> {
     fn ask_color_for(&mut self, value: &str, new_prompt: &mut String) {
-        let _ = self.write(b"What color you would like to use for this variable?:\n");
+        let _ = self.write(b"Choose a color for the variable:\n");
+        for &color in COLORS.keys {
+            let _ = write!(self, "{}, ", color);
+        }
+        let _ = writeln!(self, "or default");
         let _ = self.flush();
         let mut input = String::new();
         let _ = io::stdin().read_line(&mut input);
-        new_prompt.push_str(&input);
+        new_prompt.push_str(&["${c::", input.trim(), "}"].concat());
         new_prompt.push_str(value);
     }
 }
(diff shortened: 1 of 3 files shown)


stratact:
About the colors part, could I also have a preset of color values, such as `blue`, `green`, `red`, `yellow` etc for the end user to choose from if they don't want to use hex because it might be too hard for them to do. I'd say the hexdecimal option is more of a power user option.

mmstick:
@stratact You can display the list from the colors module.

stratact:
Ah fantastic

mmstick:
(and hexadecimal should be simple since it's literally just `0xRRGGBB`

stratact:
I'm not excluding the hexadecimal idea, it's just that there should be some options that are easier to use and get started with as users become more familiar with ion.

src/lib/builtins/prompt.rs
```diff
@@ -43,16 +48,25 @@ pub(crate) fn prompt(args: &[String], shell: &mut Shell) -> Result<(), String> {
...
+                            Some('c') => { 
+                                match input_iter.next() {
+                                    Some('1') => stdout.ask_color_for("${PWD}", &mut new_prompt),
+                                    Some('2') => stdout.ask_color_for("${SWD}", &mut new_prompt),
+                                    _ => continue,
+                                }
+                            }
+                            Some('d') => { 
+                                unimplemented!();
+                                stdout.ask_color_for(r#"$(top -bn1 | grep "Cpu(s)" | sed "s/.*, *\([0-9.]*\)%* id.*/\1/" | awk '{print 100 - $1"%"}')"#, &mut new_prompt);
```
stratact: I need help figuring out to do an Ionic way of doing `$(top -bn1 | grep "Cpu(s)" | sed "s/.*, *\([0-9.]*\)%* id.*/\1/" | awk '{print 100 - $1"%"}')`.

stratact:
Closing for now, gonna work on this in our GitLab server instead.

FriedPandaFries:
Hello, in ion's current state would it be possible to use something like this <a href="https://github.com/jD91mZM2/powerline-rs">Rust rewrite of Powerline</a>?

zen3ger:
@FriedPandaFries it should. In `~/.config/ion/initrc` define:

```rust
fn PROMPT
    powerline-rs --shell bare $?
end
```

**NOTE**: The redox project moved to [gitlab](https://gitlab.redox-os.org/redox-os/ion), so for further questions either visit the forum/chat or the gitlab issues!

Issue #755, opened 2018-06-02
Ion parsing bug when dealing with string expansion


stratact:
**Reproduction**: `echo $(let free=[@(free -h)]; echo @free[6]@free[8]/@free[7])`.

**Expected behavior**: Should output `Mem:4Gi/32Gi`.

**Actual behavior**: Outputs `]; echo /)` instead.

**Build information**:
`rustc 1.28.0-dev (a2c4d4e2f 2018-05-28)

ion 1.0.0-alpha (x86_64-unknown-linux-gnu)
rev e3bd8ce699d9c9904687ffaefc6ca1fe105d9788

**Blocking/related**: blocks me from doing the Memory usage option for Ion's `prompt` builtin command: #754

PR #756, opened 2018-06-02
Simplify the paren level code in StatementSplitter

stratact:
While this doesn't solve #755, it fixes a bug where if you have an array paren expansion inside another subshell paren expansion, the statement splitter will only decrement the `p_level` and never `ap_level`.

### DIFF
--- a/src/lib/parser/statement/splitter.rs
+++ b/src/lib/parser/statement/splitter.rs
@@ -85,9 +85,7 @@ pub(crate) struct StatementSplitter<'a> {
     read:             usize,
     start:            usize,
     flags:            Flags,
-    a_level:          u8,
-    ap_level:         u8,
-    p_level:          u8,
+    paren_level:      u8,
     brace_level:      u8,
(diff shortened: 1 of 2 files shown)


mmstick:
Can you add a unit test for the bug it fixes?

stratact:
Just for documenting the reason, since I've already mentioned this on Mattermost, but since stuff there can get lost, I'll put it here for future reference:

It hasn't really solved any bug except when you do this
```
             $(             @(              ) )
^             ^              ^              ^
|             |              |              |
|             |              |              |
p_level = 0   p_level = 1    p_level = 1    p_level = 0
ap_level = 0  ap_level = 0   ap_level = 1   ap_level = 1
```
Instead I made it like this:
```
                $(              @(               )                )
^                ^               ^               ^                ^ 
|                |               |               |                |
|                |               |               |                |
paren_level = 0  paren_level = 1 paren_level = 2 paren_level = 1  paren_level = 0
```
to act like a normal stack

PR #760, opened 2018-06-04
Fix for #682

### LINKED ISSUE #682: Redirection of stdin results in panic on unrwrap()
For example `echo ls | ./target/debug/ion` results in:
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 25, kind: Other, message: "Inappropriate ioctl for device" }', libcore/result.rs:916:5

The problematic unwrap is at src/lib/shell/binary/readln.rs:30.
The ultimate problem is that liner tries to use a raw input mode, but this will fail for a redirected stdin.

Since I'm not sure where this problem should be fixed, I opened a similar issue at https://github.com/MovingtoMars/liner/issues/77
See also: https://github.com/ticki/termion/issues/39

stratact:
This fixes #682 but it hangs when doing redirection inside an Ion session, ~~probably likely due to the looping mentioned in https://github.com/redox-os/liner/pull/2~~

~~This PR also requires Ion to migrate to our Liner fork with https://github.com/redox-os/liner/pull/2 merged.~~

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -56,6 +56,7 @@ fn main() {
         shell.execute_interactive();
     } else {
         let reader = BufReader::new(stdin());
-        shell.terminate_script_quotes(reader.lines().filter_map(|line| line.ok())); 
+        let status = shell.terminate_script_quotes(reader.lines().filter_map(|line| line.ok())); 
+        shell.exit(status);
     } 
 }
--- a/src/main.rs
+++ b/src/main.rs
@@ -9,11 +9,16 @@ use std::{
 };
 
 fn main() {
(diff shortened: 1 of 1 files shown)


mmstick:
It shouldn't be caused by liner, since liner will never be initialized.

stratact:
Alright I'll try narrowing it down where the hanging happens and share what I find.

stratact:
Okay, the reason it hangs is that it creates a zombie child process.

stratact:
@mmstick I'm starting to think it is a Liner issue, because I'm executing the redirection inside the interactive Ion session.

stratact:
Nice work! :+1:

Issue #693, opened 2018-02-16
panic for let array] = "zero"


haraldh:
**Reproduction**: [describe how you are able to reproduce ("trigger") this bug/issue.]
```
$ RUST_BACKTRACE=1 cargo run
   Compiling ion-shell v1.0.0-alpha (file:///home/harald/git/ion)
    Finished dev [unoptimized + debuginfo] target(s) in 3.81 secs
     Running `target/debug/ion`
# let array] = "zero"
thread 'main' panicked at 'attempt to subtract with overflow', src/lib/parser/quotes.rs:105:33
stack backtrace:
   0: std::sys::unix::backtrace::tracing::imp::unwind_backtrace
             at libstd/sys/unix/backtrace/tracing/gcc_s.rs:49
   1: std::sys_common::backtrace::_print
             at libstd/sys_common/backtrace.rs:71
   2: std::panicking::default_hook::{{closure}}
             at libstd/sys_common/backtrace.rs:59
             at libstd/panicking.rs:380
   3: std::panicking::default_hook
             at libstd/panicking.rs:396
   4: std::panicking::rust_panic_with_hook
             at libstd/panicking.rs:576
   5: std::panicking::begin_panic
             at libstd/panicking.rs:537
   6: std::panicking::begin_panic_fmt
             at libstd/panicking.rs:521
   7: rust_begin_unwind
             at libstd/panicking.rs:497
   8: core::panicking::panic_fmt
             at libcore/panicking.rs:71
   9: core::panicking::panic
             at libcore/panicking.rs:51
  10: ion_shell::parser::quotes::Terminator::is_terminated
             at src/lib/parser/quotes.rs:105
  11: ion_shell::shell::binary::terminate::terminate_quotes
             at src/lib/shell/binary/terminate.rs:57
  12: <ion_shell::shell::Shell as ion_shell::shell::binary::Binary>::terminate_quotes
             at src/lib/shell/binary/mod.rs:79
  13: <ion_shell::shell::Shell as ion_shell::shell::binary::Binary>::execute_interactive
             at src/lib/shell/binary/mod.rs:147
  14: ion::main
             at src/main.rs:58
  15: std::rt::lang_start::{{closure}}
             at /checkout/src/libstd/rt.rs:74
  16: std::panicking::try::do_call
             at libstd/rt.rs:59
             at libstd/panicking.rs:479
  17: __rust_maybe_catch_panic
             at libpanic_unwind/lib.rs:102
[...]

PR #763, opened 2018-06-05
Fix for #693

stratact:


src/lib/parser/quotes.rs
```diff
@@ -81,9 +82,14 @@ impl Terminator {
...
                             b']' if !self.flags.intersects(Flags::DQUOTE | Flags::SQUOTE) => {
-                                self.array -= 1;
-                                if self.array == 0 {
+                                if self.array > 0 {
+                                    self.array -= 1;
+                                } else if self.array == 0 && self.flags.contains(Flags::ARRAY) {
                                     self.flags -= Flags::ARRAY
+                                } else {
+                                    instance |= Flags::ERROR;
+                                    eprintln!("ion: syntax error: ion: syntax error: extra right bracket(s)");
```
mmstick: double syntax error

stratact: What do you mean exactly?

PR #765, opened 2018-06-05
Fix for #695

### LINKED ISSUE #695: Unexpected end of script when nesting if inside match
**Reproduction**: Run the following piece of code
```
match "foo"
    case "foo"
        if true
            echo "This is executed"
        end
end
```

**Expected behavior**: "This is executed" gets printed (it does) and ion exits successfully.

**Actual behavior**: "This is executed" gets printed and ion exits with "ion: unexpected end of script: expected end block for `Default`" and 1 as an exit code

**Build information**: 
Ion built with `rustc 1.25.0-nightly (3ec5a99aa 2018-02-14)`,

 `ion 1.0.0-alpha (x86_64-apple-darwin)
rev 22b3efc6eb29020b452c44ad057c394951f6b982`

**Misc**: This issue also occurs when trying to run `ion setup.ion build docs`.

Note: I just started looking into ion, so this might just be me being unable to grasp the syntax

stratact:
I'm not proud of this fix, but it was the only way.

Closes #695

### DIFF
--- a/src/lib/shell/flow.rs
+++ b/src/lib/shell/flow.rs
@@ -339,6 +339,8 @@ impl FlowLogic for Shell {
         else_if: Vec<ElseIf>,
         failure: Vec<Statement>,
     ) -> Condition {
+        // In cases when the level is not 0, just reset it to 0
+        self.flow_control.level = 0;
         let first_condition = iter::once((expression, success));
         let else_conditions = else_if
             .into_iter()


mmstick:
I think we should wait for a permanent fix to the issue. This will fix the example in 695, but there are many scenarios where this will cause unintended side effects (multiple if statements)

stratact:
So the primary reason this bug happens is that after all the parsing is done and when it lands [here](https://github.com/redox-os/ion/blob/master/src/lib/shell/flow.rs#L458), which increases `self.flow_control.level` to `1` but it never decrements it back to `0`.

So either `collect_if()` has to decrement it but it can't because the statement matches to `Statement::Pipeline`, or `self.execute_if()` which is the fix I made.

So I don't know how to properly fix this :man_shrugging:

(and yes I tried matching `statement` to `Statement::Pipeline` and decrementing it by 1, but that fails the integration tests)

I mean I do admit my fix does feel hack-ish but I don't know any other way.

zen3ger:
I had an idea of how I would go about it, but it would've been a larger rewrite of the current control-flow parsing/execution, but I haven't got the time and Rust knowledge to implement it back then.

The idea is that `FlowControl` would have a stack, where new frame is pushed when a block opens, and on each block ending keyword the top frame is popped and merged back to the previous frame `statements` section.

```lua
match $var                     #  -- push Statement::Match
  case 1                       #      |-- push Statement::Case 
    if true                    #      |    |-- push Statement::If
      echo "A"                 #      |    |   |-- insert into `If.success`
    else                       #      |    |   |
       echo "Naaah..."         #      |    |   |-- insert into `If.failure`
    end                        #      |    |   `-- pop and merge back to `Case.0.statements`
                               #      |    `-- pop and merge back to `Match.cases`
  case 2                       #      |-- push Statement::Case
    echo "B"                   #      |   |-- insert into `Case.0.statements`
                               #      |   `-- pop and merge back to `Match.cases`
end                            #      `-- pop, stack is empty => execute parsed block
```
I think it could be usefull for reporting errors too, where frames can have the line no. where they were pushed, so when invalid token comes up it could be reported when and what block started. I'm very interested about your opinions about this solution.

stratact:
Well I think @mmstick would want to avoid using a `Vec` to emulate a stack and rely on recursion mostly for performance reasons, but I'll let him decide on your suggestion.

mmstick:
That's basically how it works at the moment. It is possible to get line information if needed.

mmstick:
@stratact A `Vec` is used to collect statements for each block. Blocks are collected via recursive descension.

stratact:
Okay after a lot of thought and debugging, I think @zen3ger has the better approach, because the current flow control logic is flawed. After collecting all the statements and executing them, they cannot increase the `flow_control.level` from `0`.

This is going to require a refactor much bigger than I thought, which I'm not up for nor do I understand the shell logic fully. So I'm closing this.

@mmstick please change the label in #695 to at least `A-level` since this is not a `good first issue` issue

mmstick:
@stratact It's basically the same approach as what the flow control logic is doing right now. The `level` is used to tell the flow control logic when to stop collecting statements from the input iterator. This is likely just a matter of something not incrementing the level when it should have.

stratact:
@mmstick Let me illustrate what the problem is when I run the example in #695 to show how the level jumps:
```
`Shell.flow_control.level` Progression

Old Value  New Value  Modified From
---------  ---------  ------------------
0          1          execute_toplevel()
1          2          collect_cases()
2          1          collect_cases()
1          0          collect_cases()
0          1          execute_statement() <-- WTF?

```

stratact:
I mean, I'm sorry, it's just that when `execute_statement()` happens as illustrated above, it increases the level back to one, everything else seems to function normally going from `0` -> `1` -> `2` -> `1` -> `0`. This bug has been frustrating.

zen3ger:
@stratact I've looked through this many times and what I found was that the shell level is being set properly. What was interesting is that
```lua
match "a"
   case "a"
     if true
       echo OK
     end
 end
``` 
and
```lua
match "a"
  case "a"
     if true
     end
     echo OK
 end
``` 
IS represented the same way.  The echo never been pushed into the if body but into the statements of case.

PR #775, opened 2018-06-09
Have `WordIterator` not split escape characters (Fixes #437)

### LINKED ISSUE #437: wildcard expansion after escaping
```
# touch 'f x' 'f y'
# rm f\ *
rm: cannot remove ' *': No such file or directory
rm: cannot remove 'f': No such file or directory
# rm f*
```

stratact:
Although this solves the problem, it fails a variety amount of integration tests, which probably need to be updated for this different logic. I would like some guidance on how to fix them first.

Closes #437

### DIFF
--- a/src/lib/parser/shell_expand/words/tests.rs
+++ b/src/lib/parser/shell_expand/words/tests.rs
@@ -47,12 +47,11 @@ fn string_method() {
 
 #[test]
 fn escape_with_backslash() {
-    let input = "\\$FOO\\$BAR \\$FOO";
+    let input = r#"\$FOO\$BAR \$FOO"#;
     let expected = vec![
-        WordToken::Normal("$FOO", false, false),
-        WordToken::Normal("$BAR", false, false),
+        WordToken::Normal("$FOO$BAR".into(), false, false),
         WordToken::Whitespace(" "),
-        WordToken::Normal("$FOO", false, false),
+        WordToken::Normal("$FOO".into(), false, false),
     ];
     compare(input, expected);
 }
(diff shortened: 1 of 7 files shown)


src/lib/parser/shell_expand/words/tests.rs
```diff
@@ -210,26 +210,26 @@ fn test_words() {
...
+//        WordToken::Normal("foo", false, false),
+//        WordToken::Normal("(", false, false),
+//        WordToken::Normal(")", false, false),
+//        WordToken::Whitespace(" "),
+//        WordToken::Normal("bar", false, false),
+//        WordToken::Normal("(", false, false),
+//        WordToken::Normal(")", false, false),
+//    ];
+//    compare(input, expected);
+//}
```
mmstick: This should be passing. Type `echo foo\(\) bar(\)` into the prompt. You should get  `foo() bar()`

stratact: All I get is this
```
# echo foo\(\) bar(\)
ion: syntax error: '(' at position 17 is out of place
```

src/lib/parser/shell_expand/words/tests.rs
```diff
@@ -210,20 +210,20 @@ fn test_words() {
...
-//    ];
-//    compare(input, expected);
-//}
+#[test]
+fn test_multiple_escapes() {
+    let input = "foo\\(\\) bar\\(\\)";
+    let expected = vec![
+        WordToken::Normal("foo".into(), false, false),
+        WordToken::Normal("(".into(), false, false),
+        WordToken::Normal(")".into(), false, false),
```
mmstick: This should be `WordToken::Normal("foo()", ...),`

Issue #397, opened 2017-07-11
`sudo` Tab Completions Should Complete Commands


xTibor:
The `sudo` command should have command completions. At the moment the tab completions suggests files and directories for this command.

**Build information**:
`git rev-parse HEAD`: eb13ba645bad3b311843f38f84455ce3a53fdb8d
`rustc -V`: rustc 1.20.0-nightly (720c596ec 2017-07-08)

drosseau:
I've been working on this a bit and I'm wondering if we want to try to support importing already written bash/zsh style custom completions? It'd make switching to ion a lot smoother for people, but it might be a lot of work since their syntax is based on their languages.

mmstick:
We could do as Fish does and generate completions from man pages detected in the system.

theduke:
For reference, here is the Python script fish uses to generate completion info: https://github.com/fish-shell/fish-shell/blob/master/share/tools/create_manpage_completions.py

BafDyce:
Whats the current plan for implementing this?

Tbh, not having auto-completion is the only thing that really prevents me from fully switching to ion as my default shell.

mmstick:
No plans at the moment (at least not from me). I think we should study what Fish is doing, and then come up with a plan that is at least equivalent to or superior to what they are doing.

theduke:
For me, three steps make sense:

* Make user-defined completions possible with shell functions / external commands
* Provide a compatibility layer that supports bash or zsh completion support (by just executing them and using the output probably)
* Provide auto-generated completions based on man pages

To me, step 1 is the most essential, since then I can just code up the completions I need before support gets more advanced, and the other steps kind of require it.

I'm willing to take a stab at this, but I'd appreciate some pointers on how/where/how that should be done.

BafDyce:
I'm with you that step one is essential. However, I think the best solution would be to store the required information in a config file (which are then loaded into an apropriate memory structure on ions shartup and/or reloaded if a special builtin-command is executed). With this config loaded, ion can then just look in this data structure to see what to suggest for tab-completion.

I think at least the following (and probably many more) properties should be configurable/available:
- aliases - a list of application names that config file is for (since there are quite a few applications which also have shortcuts, e.g. `netcat` and `nc`)
- subcommands - a list of available sub-commands (e.g. `pull`, `push`, `commit`, etc. for `git`). Each with a distinct set of parameters, however they might also share some
- parameter attributes
    - files
    - files with a specific extension (e.g. `.rs`)
[...]

PR #777, opened 2018-06-10
WIP: Tab completions

BafDyce:
**IMPORTANT**
This is sorta, kinda WIP and NOT intended to be merged as it is!
The code still needs a few tweaks, and the commits are not meaningful on their own (I just committed all wip-stuff at once when I made breaks). This is more like a proposal and a basis for discussion.

**Problem**:
This PR adds support for tab completions. See  #397

**Solution**:
The general idea: For each application, there is a user-friendly config file (TOML-format) which defines different types of command line parameters, etc. Currently supported features:
- boolean flags (e.g. `--version`)
- paths
- files with specific file extensions
- numbers
- key=value
- string (aka "everything else that takes custom values") with support for options (predefined list of possible values)

If an application has no completion-config, the default file file completer is used.

- New dependencies: toml, serde
- On startup (in `execute_interactive`), available completion-configs are parsed and saved. **For now, this is hardcoded**, as I'm unsure where we should save such files.
- A new completer (`IonCmdCompleter`) was created, similar to the existing `IonFileCompleter`
- When the user hits tab twice and the application (first word on the command line) has a completion defined, it is examined for possible matches.
- I've added a base ("empty") config file as well as two (unfinished) configs for `cargo` and `git`.

**Drawbacks**:
- I don't know if my code structure/architecture is fine, as I figured out how liner works as I worked on this, so the code may be restructured more cleanly?
- Also, I had to add an unsafe (https://github.com/BafDyce/ion/blob/adb13743757078f021614e89a8820958ea15203e/src/lib/shell/completer.rs#L71) to satisfy the borrow checker. Don't know if this can be avoided by restructuring.
- Aliases (different names for the same application) are not handled/implemented yet.

**TODOs**:
- Discuss whether such a config-file based solution is wanted
- Add support for "dynamic" parameters (list of completions generated at runtime by executing an `ion` script?)
- Add support for "scheme" parameters
- Decide where to save the config files
    - Add hook for installation script
    - Adapt initialization code accordingly
- Fix issue: auto-complete does not work if the cursor is on the right edge of a completed parameter (no space is inserted afterwards)
- Cleanup the code, Add comments
- Rebase onto current master and fix merge conflicts
- We might also want to add a new builtin for (re)loading completions at runtime
- In a next step, we also should implement parsers to import completions of other shells/from man-pages.
- Test on Redox (currently only on Linux tested)
- Add a bunch of completion configs.

**Fixes**: #397

**State**: WIP

**Blocking/related**: None that I know of.

**Other**:
- Is the direction of this PR/proposal good? Or would you rather do it completely script-based (like other shells do)? I think the major advantage of a config-based approach is, that it's easier to set up a new completion ruleset.
- This is WIP and not ready to merge, however I wanted to have you have a look at it before I invest more time in it. Doesn't make sense to continue my efforts if you don't even want such a solution.

I'm open for feedback, however I'll also be busy in the next days.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -171,8 +171,11 @@ dependencies = [
  "rand 0.4.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "redox_syscall 0.1.38 (registry+https://github.com/rust-lang/crates.io-index)",
  "regex 1.0.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.61 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_derive 1.0.66 (registry+https://github.com/rust-lang/crates.io-index)",
  "smallstring 0.1.2 (registry+https://github.com/rust-lang/crates.io-index)",
  "smallvec 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "toml 0.4.6 (registry+https://github.com/rust-lang/crates.io-index)",
  "unicode-segmentation 1.2.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "users 0.6.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "version_check 0.1.3 (registry+https://github.com/rust-lang/crates.io-index)",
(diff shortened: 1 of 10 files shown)


mmstick:
@BafDyce You may want to PR this [here](https://gitlab.redox-os.org/redox-os/ion).

mmstick:
I'm wondering if [RON](https://github.com/ron-rs/ron/) might be a better format than TOML for this?

mmstick:
Thanks for looking into this. It's massive undertaking. This looks like it's going in the right direction. We definitely need the ability to parse completions from man pages.

BafDyce:
@mmstick: Sure, is it possible to create an account there or should I use the 'login via github' option?

I can look into RON, I just used TOML because it's used often in the Rust ecosystem. Also I agree to you that we need a way to parse man pages.

mmstick:
@BafDyce Yes, it'd be good to have the PR there. RON might work a bit better in this case. TOML syntax is a bit awkward for the complex rules like we may need here.

PR #782, opened 2021-10-07
proposed grammar, punctuation, and word choice changes for introduction

brandondrew:
Ion looks amazing!  As I was reading the introduction, I noticed some things I felt could be improved and thought I would make a very small contribution.

I can explain any of the changes, if the rationale is not clear.


brandondrew:
> New users must create an account and then send an email to info@redox-os.org requesting approval.

Can I reasonably expect approval?

chalst:
I would guess so. You can cite this thread in your mail, which might make the approval process go faster.

hardBSDk:
@brandondrew @chalst We had many problems with email approval in the past because the email went offline many times.

Thus we moved to a room on Matrix where you post your GitLab username, it fixed all account approval problems.

PR #784, opened 2022-07-09
Update README.md

sarashino:
it seems like just a dead link.


hardBSDk:
If this PR is still necessary repost this PR in the [GitLab repository](https://gitlab.redox-os.org/redox-os/ion), this GitHub repository is just a mirror of GitLab and we don't accept PRs here.

PR #787, opened 2023-04-15
Fine-tune the atomic::ordering for all the atomics

wang384670111:
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L30
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L46
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L51
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L57
Here grab is just a signal and doesn't synchronize with other variables. Therefore, Relaxed can be used in both single-threaded and multi-threaded environments.

https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L39
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L52
Here status is just a signal and doesn't synchronize with other variables. Therefore, Relaxed can be used in both single-threaded and multi-threaded environments.
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L34
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L47
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L53
Here reply is used to synchronize the status variable. Therefore, the read operation of reply needs to see the changes of the status variable. Using Ordering::SeqCst is unnecessary and may impact program performance, I think just Acquire/Release needs to be used here to ensure the correctness of the program.
https://github.com/redox-os/ion/blob/0fe51451ddda3797444df06d39519a39e3a31e36/src/lib/shell/pipe_exec/foreground.rs#L35
Here reply is just a signal and doesn't synchronize with other variables. Therefore, Relaxed can be used in both single-threaded and multi-threaded environments.

### DIFF
--- a/src/lib/shell/mod.rs
+++ b/src/lib/shell/mod.rs
@@ -180,7 +180,7 @@ impl<'a> Shell<'a> {
                 _ => unreachable!(),
             };
 
-            signals::PENDING.store(signal as usize, Ordering::SeqCst);
+            signals::PENDING.store(signal as usize, Ordering::Relaxed);
         }
 
         unsafe {
--- a/src/lib/shell/pipe_exec/foreground.rs
+++ b/src/lib/shell/pipe_exec/foreground.rs
@@ -27,34 +27,34 @@ pub struct Signals {
 
 impl Signals {
     pub fn was_grabbed(&self, pid: Pid) -> bool {
-        self.grab.load(Ordering::SeqCst) == pid.as_raw() as usize
+        self.grab.load(Ordering::Relaxed) == pid.as_raw() as usize
     }
 
     pub fn was_processed(&self) -> Option<BackgroundResult> {
-        let reply = self.reply.load(Ordering::SeqCst) as u8;
-        self.reply.store(0, Ordering::SeqCst);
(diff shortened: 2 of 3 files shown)


hardBSDk:
@wang384670111 If this PR is still necessary repost this PR in our [GitLab repository](https://gitlab.redox-os.org/redox-os/kernel), this is just a mirror of GitLab and we don't accept PRs here.

PR #788, opened 2023-12-24
Fix `set -e` for conditionals

riyad:
When porting scripts to Ion I noticed that using `set -e` scripts would exit early even when conditions fail. This is in contrast to common shells like  Bash which has a [list of exceptions](https://www.gnu.org/software/bash/manual/html_node/The-Set-Builtin.html) exactly for this case.

### Approach

In this PR I've tried to tackle the problem in the most naive way:
- keep creating `PipelineError::EarlyExit` to "capture intent"
- prevent result/error propagation for `PipelineError::EarlyExit` in certain locations 
- re-raise if necessary

I've also added specific test cases to test proper behavior and copied tests for conditionals to also test with `set -e`.

### Known Problems
- I've currently no idea how to ignore failing subshells in conditions ([ see `if ! $(false | true) ...` below](https://github.com/redox-os/ion/compare/master...riyad:ion:fix-set-e-for-conditions?expand=1#diff-ac46223d85d954bd219abcca3da82b8e6d9f62b7a5f013d77c7189215547d5f5R3))

Please feel free to rewrite my PR or find a better approach to this issue.

Note: `master` branch is currently out of sync beween Redox-OS Gitlab and Github.


hardBSDk:
@riyad If this PR is still necessary repost this PR in our [GitLab repository](https://gitlab.redox-os.org/redox-os/kernel), this is just a mirror of GitLab and we don't accept PRs here.

PR #789, opened 2023-12-24
Add set -x

riyad:
This adds `set -x` (used e.g. in Bash) for printing every line before executing it.

The code is the same as for `debug on`.
The `set`-variant seems useful for porting scripts, but supporting it may not be desired, because the feature exists already.

Alternatively `set -x` could be an undocumented option pointing users to use `debug on`.

Note: `master` branch is currently out of sync beween Redox-OS Gitlab and Github.


hardBSDk:
@riyad If this PR is still necessary repost this PR in our [GitLab repository](https://gitlab.redox-os.org/redox-os/kernel), this is just a mirror of GitLab and we don't accept PRs here.

